PR #2, opened 2022-08-14
[Tin En Hao] iP

tinenhao:
_DukePro frees your mind of having to remember things you need to do. It's,_

1. text-based
2. easy to learn
3. ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/tinenhao/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.
5. And it is FREE!

Features:

- [x] Managing tasks
- [ ]  Managing deadlines (coming soon)
- [ ]  Reminders (coming soon)


src/main/java/Commands/Command.java
```diff
@@ -0,0 +1,36 @@
...
+ * Parent class of all Duke Commands
+ */
+public abstract class Command {
+    private boolean isExit;
+
+    public Command() {
+        this.isExit = false;
+    }
+
+    public void setExit() {
```
deeyonn, line 18: What is the rationale for allowing `12: private boolean isExit` to be set by `12: public void setExit()`

Another alternative would be to default `12: public void setExit()` to false,
```java
// abstract class Command
public boolean isExit() {
    return false;
}
```

and override it in an exit command.

```java
// class ExitCommand
@Override
public boolean isExit() {
    return true;
}
```

deeyonn (commented):
Looks great to me! 😄

src/main/java/Constants.java
```diff
@@ -0,0 +1,23 @@
...
+    public static final String deleteTask = "Noted. I've removed this task: ";
+
+    public static String numTasks(int i) {
+        return String.format("Now you have %d tasks in the list.", i);
+    }
+
+    public static final String invalid = "OOPS!!! I'm sorry, but I don't know what that means :-(";
+
+    public static final String invalidInput = "Something is missing form your input";
+}
```
ryanlml: I like how you've abstracted all the error messages into its own class! It makes your job easier if you were to change the error messages in the future. However, according to the coding standard, 

`Constant names must be all uppercase using underscore to separate words.`

so perhaps you may consider renaming them!

src/main/java/Duke/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+            checkInputLength(parse0,2);
+            return new PriorityCommand(parseInt(parse0[0]), parse0[1]);
+        case "todo":
+            checkInputLength(parse,2);
+            assert parse.length == 2: "Error in input for todo";
+            return new ToDoCommand(parse[1]);
+        case "deadline":
+            checkInputLength(parse,2);
+            assert parse.length == 2: "Error in input for deadline";
+            String[] parse2 = parse[1].split(" /by ");
```
ryanlml, line 63: Variable names here like `parse2` can perhaps be more precise! Maybe name them as `taskType`, `taskName` or `deadlineDate` so that it is easier for the reader to keep track of what the strings are!

src/main/java/Duke/Parser.java
```diff
@@ -0,0 +1,88 @@
...
+        case "bye":
+            checkInputLength(parse, 1);
+            assert parse.length == 1: "Error in input for bye";
+            return new ExitCommand();
+        case "list":
+            checkInputLength(parse, 1);
+            assert parse.length == 1: "Error in input for list";
+            return new ListCommand();
+        case "mark":
+            checkInputLength(parse,2);
```
ryanlml, line 35: Tiny issue, but note the spacing that is required here: 
```suggestion
            checkInputLength(parse, 2);
```

ryanlml (commented):
Overall your code is easy to read and pleasant on the eyes! I especially liked how the command class has been extracted. It is not easy but you did it beautifully! I noted a few minor coding violations but they should be an easy fix :)

PR #7, opened 2022-08-17
[Dexter Sim] iP

dexter-sim:
# Duke

Duke is a task tracking software that is
- text-based
- easy to learn

All you need to do is
1. download it from [here](https://github.com/dexter-sim/ip).
2. double click it.
3. add tasks.

And it is **FREE**!

Features
- [x] Manage tasks
- [x] Manage deadlines
- [ ] Reminders (coming soon) 
---

```java
public class Main {
    }
}
```


src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,75 @@
...
+        return isDone ? "X" : " ";
+    }
+
+    /**
+     * Returns the task in a string format to be saved in a local file.
+     *
+     * @return A string corresponding to the task.
+     */
+    public String stringify() {
+        return String.format("%d | %s", isDone ? 1 : 0, description);
```
Tan-Jia-Rong, line 68: Can afford to improve code quality  by avoiding complicated expressions
Personally, I feel that it would be better to compute the predicate before putting it into the return statement for clarity
```suggestion
        int doneStatus = isDone ? 1 : 0;
        return String.format("%d | %s", doneStatus, description);
```

Tan-Jia-Rong: Can afford to improve code quality  by avoiding complicated expressions
Personally, I feel that it would be better to compute the predicate before putting it into the return statement for clarity
```suggestion
        int doneStatus = isDone ? 1 : 0;
        return String.format("%d | %s", doneStatus, description);
```

Tan-Jia-Rong (changes requested):
LGTM! :+1:

Just some nits to pick but overall, I found your code to be quite readable and the association between classes are also quite logical
I've only managed to find a minor details that I felt would help to improve the code quality. However aside from that all is well 😄

src/main/java/duke/task/Task.java
dexter-sim: Thank you for the suggestion. I like your suggestion and I have committed your suggested change.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+public class Parser {
+    /**
+     * Returns a command based on the specified user input.
+     *
+     * @param userInput Represents the user input into the system.
+     * @return A command based on the user input.
+     * @throws DukeException If user input is an invalid string command.
+     */
+    public static Command parseInput(String userInput) throws DukeException {
+        String[] split = userInput.split(" ");
```
njxue, line 33: Do you think it's better if you add an additional line `String command = split[0]` here? That way, you can use `command.equals(xxx)` rather than `split[0].equals(xxx)`, which I think is more readable 🤣 .

njxue: Do you think it's better if you add an additional line `String command = split[0]` here? That way, you can use `command.equals(xxx)` rather than `split[0].equals(xxx)`, which I think is more readable 🤣 .

njxue (commented):
LGTM!

src/main/java/duke/Parser.java
dexter-sim: It is a good idea. I will implement your suggestion. Thank you for your review!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,107 @@
...
+        if (userInput.equals("welcome") && split.length == 1) {
```
24Donovan24: Do you think it will be neater to implement the conditionals as a switch case rather than multiple ifs statements?

src/main/java/duke/Ui.java
```diff
@@ -20,109 +18,84 @@ public class Ui {
...
-
-    /**
-     * Creates ui object to handle input and output.
-     */
-    public Ui() {
-        scanner = new Scanner(System.in);
-    }
-
     /**
      * Prints Duke logo and greeting message.
```
24Donovan24: JavaDocs can include the return type for some of the methods here

24Donovan24 (commented):
LGTM! Just some minor nits to fix!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,76 @@
...
+        this.isDone = isDone;
+    }
+
+    /**
+     * Returns a boolean if the description contains specified keyword.
+     *
+     * @param keyword A string to be checked.
+     * @return If description contains specified keyword.
+     */
+    public boolean match(String keyword) {
```
turretDive, line 43: Maybe it would be good to named Boolean methods to sound like booleans? Such as canMatch() or isMatch().

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,132 @@
...
+    public List<Task> getTasks() {
+        return tasks;
+    }
+
+    /**
+     * Returns the size of the list of tasks.
+     *
+     * @return Size of the list.
+     */
+    public int size() {
```
turretDive: Maybe it would be good to use verbs for Actions. Such as getSize().

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,51 @@
...
+        super(description, isDone);
+        this.by = by;
+    }
+
+    private String printDate() {
+        return DateTimeFormatter.ofPattern("MMM dd yyyy").format(by);
+    }
+
+    @Override
+    public String stringify() {
```
turretDive, line 48: Adding a comment here on the difference between stringify() and toString() such as the difference in uses will explain why there is a need for another similar function 👍

turretDive (commented):
LGTM! Overall just come nits found in naming convention and documentation etc. I really liked how clean the overall code looked!

PR #8, opened 2022-08-17
[Dominic Berzin] iP

domoberzin:
# Duke ChatBot!
A personal task and deadline tracking tool! It's,

- text-based
- simple and easy to use
- runtime is not O(2^n) (aka its kinda fast i guess)

How do you get started?

1. Download it [here](https://github.com/domoberzin/ip/releases).
2. Double-click it.
3. Add your tasks, deadlines or events.
4. ???
5. Profit.

Price: $0

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon) 

If you're a Java programmer, you can practice Java with it too! Here's the main method:

```java
    Duke chatBotInstance = new Duke();
    chatBotInstance.run();
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Reads and handles user input until parser returns boolean to indicate end.
+     *
+     */
+    private void loopInputRead() {
+        Parser parser = new Parser(new Scanner(System.in));
+        while (true) {
+            try {
+                boolean complete = parser.handleInput();
```
therealdaofu: Since it is a boolean, maybe you could rename it to sound like a boolean, maybe isComplete would be better?

src/main/java/duke/gui/DialogBox.java
```diff
@@ -0,0 +1,74 @@
...
+        try {
+            FXMLLoader fxmlLoader = new FXMLLoader(MainWindow.class.getResource("/view/DialogBox.fxml"));
+            fxmlLoader.setController(this);
+            fxmlLoader.setRoot(this);
+            fxmlLoader.load();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        Circle shape = new Circle(50, 50, 50);
```
therealdaofu, line 42: maybe you could just name it circle instead of shape?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+import duke.exceptions.DukeException;
+import duke.tasklist.TaskList;
+
+/**
+ * Parser to handle user input.
+ */
+public class Parser {
+
+    /** Scanner to take in user input. */
+    private Scanner in;
```
therealdaofu: perhaps it could be clearer if you named it something like input instead?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+                tasklist.addTask(next);
+                break;
+            case "find":
+                tasklist.findWithFilter(next[1]);
+                break;
+            case "delete":
+                tasklist.delete(next[1]);
+                break;
+            default:
+                throw new DukeException("Invalid command");
```
tsammeow: Maybe these could be abstracted into a `Command` class so handling different types of commands would be easier?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+        return false;
+    }
+
+    /**
+     * Handles user input and calls methods accordingly.
+     *
+     * @return Boolean to indicate end command has been sent.
+     * @throws DukeException In the event that the command is not recognised.
+     */
+    public String handleGuiInput(String input) throws DukeException {
```
tsammeow: Instead of having two methods that do almost the same thing that differ only based on the input and output method, how about abstracting the I/O into a different class(es)? Then you could have a single method read from that class' input method and output to that class' output method.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,91 @@
...
+
+    }
+
+    /**
+     * Gets response from parser based on user input.
+     * @param input User input.
+     */
+    public String getResponse(String input) {
+        try {
+            Parser parser = new Parser(new Scanner(System.in));
```
tsammeow: Maybe `Parser` could be refactored to not need a `Scanner` any more? GUI input is now being used, so standard input is no longer required here.

src/main/java/duke/gui/DialogBox.java
```diff
@@ -0,0 +1,74 @@
...
+import javafx.scene.control.Label;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+import javafx.scene.shape.Circle;
+
+
+
+/**
+ * An example of a custom control using FXML.
```
tsammeow, line 22: Should the Javadoc here be edited away?

src/main/java/duke/gui/DialogBox.java
```diff
@@ -0,0 +1,74 @@
...
+        Collections.reverse(tmp);
+        getChildren().setAll(tmp);
+        setAlignment(Pos.TOP_LEFT);
+    }
+
+    public static DialogBox getDukeIntro(String text, Image img) {
+        var db = new DialogBox(text, img);
+        db.flip();
+        return db;
+    }
```
tsammeow, line 63: What's the difference between this method and `getDukeDialog`?

tsammeow (commented):
Apart from some minor issues with abstraction, and some methods missing documentation, looks good to me!

PR #10, opened 2022-08-18
[Russell Saerang] iP

RussellDash332:
Stash it, get it done!
- easy to use
- easy to not use
- very ~~1337~~ COOL message, _trust me!_

Use this application by setting it up in **under one minute** (~~unless bad WiFi~~)
1. go to [this page](https://github.com/RussellDash332/ip/releases/tag/A-Jar)
2. read the instructions there
3. have fun! 🥳 

Current features include:
- [X] Support _**basic**_ operations such as add, delete, find
- [x] @RussellDash332's wishlist to support multiple tasks in one go
- [ ] Literally anything else

The application mainly uses Java, so here's what a Java code looks like.
```java
public class Main { // wow syntax highlighting!
    }
}
```
Maybe a little bit of shell script?
```sh
echo $PATH
```


src/main/java/stashy/parser/Parser.java
```diff
@@ -0,0 +1,317 @@
...
+ */
+public class Parser {
+    public static final Pattern TASK_DATA_FORMAT =
+        Pattern.compile("^\\[(?<category>[TDE])\\]"
+                + "\\[(?<isDone>[ X])\\]"
+                + " (?<description>[^\\(\\)\\n]*)\\(?"
+                + "(?<by>(by: [^\\(\\)]*)?)"
+                + "(?<at>(at: [^\\(\\)]*)?)\\)?$");
+
+    public static final String[] ACCEPTABLE_DATETIME_FORMATS = {
```
darrentde, line 40: I like how you accept multiple multiple formats for reading date time and extracted them into a constant array👍

darrentde (commented):
LGTM. Your code looks really clean and I didn't find any violations of Java coding standards. Good job 🤩

src/main/java/stashy/parser/Parser.java
RussellDash332: Thanks! I feel this is an ad-hoc solution to detect multiple date formats but it actually makes the parser more readable 😅

src/main/java/stashy/parser/Parser.java
```diff
@@ -0,0 +1,325 @@
...
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.ResolverStyle;
+import java.util.Arrays;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.IntStream;
+
+import stashy.commands.AddDeadlineCommand;
```
marcuslowhuiyu, line 12: I like how you import each class, instead of importing the whole package through *

src/main/java/stashy/data/task/Task.java
```diff
@@ -0,0 +1,71 @@
+package stashy.data.task;
+
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Represents a task.
+ */
+public abstract class Task {
+    protected static final DateTimeFormatter OUTPUT_DATETIME_PATTERN =
```
marcuslowhuiyu, line 9: I like how you use all uppercases to name your constant

src/main/java/stashy/commands/AddEventCommand.java
```diff
@@ -0,0 +1,60 @@
...
+import stashy.data.task.Event;
+import stashy.data.task.TaskList;
+import stashy.storage.Storage;
+import stashy.ui.Ui;
+
+/**
+ * A Command specifically designed to add an Event object.
+ */
+public class AddEventCommand extends Command {
+    public static final String KEYWORD = "event";
```
marcuslowhuiyu, line 13: Is there any reason for this to be publicly accessible?

src/main/java/stashy/launcher/Main.java
```diff
@@ -0,0 +1,34 @@
...
+            FXMLLoader fxmlLoader = new FXMLLoader(Main.class.getResource("/view/MainWindow.fxml"));
+            AnchorPane ap = fxmlLoader.load();
+            Scene scene = new Scene(ap);
+            stage.setScene(scene);
+            stage.setTitle("Stashy v0.3");
+            stage.setResizable(false);
+            fxmlLoader.<MainWindow>getController().setStashy(stashy);
+            stage.show();
+        } catch (IOException e) {
+            e.printStackTrace();
```
marcuslowhuiyu: I think you can considering throwing your own exception here instead of printStackTrace

RussellDash332: I agree, this is still taken from the example Main.java given in the JavaFX tutorial 😅

src/main/java/stashy/commands/AddEventCommand.java
RussellDash332: It could have been `protected`, thanks for highlighting this!

PR #22, opened 2022-08-19
[Sun Yu Ting] iP

Effixion:


src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Scanner;
+
+/**
+ * Represents the file interaction between the <code>Parser</code> object
+ * and the file stored on the hard disk of the user.
+ */
+public class Storage {
+
+    private static String filePath;
+    private File f;
```
lilythchu: Maybe you could name this variable more specifically. As said in the textbook, a name is not just for differentiation; it should explain the named entity to the reader accurately and at a sufficient level of detail.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,167 @@
...
+     *     exist in the <code>TaskList</code> object.
+     * @throws InvalidDescriptionException If the command does not have a valid description.
+     * @throws IOException If the <code>Storage</code> object cannot read or write to the file.
+     */
+    public String readInputString(String s) throws InvalidCommandException,
+            InvalidDescriptionException, IOException {
+        String[] splitCommand = s.split(" ", 2);
+        String command = splitCommand[0];
+        switch (command) {
+        case "bye":
```
lilythchu: This method is quite long. Maybe you could split it into some small methods for each case.

lilythchu (commented):
Your code looks good! I feel that your code was easy to read as it was very organized.

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,167 @@
...
+import duke.tasklist.TaskList;
+import duke.ui.Ui;
+
+/**
+ * Represents the location where the commands entered into the program are
+ * processed and how should the <code>Ui</code>, or Seaward, should respond.
+ */
+public class Parser {
+
+    private static TaskList taskList;
```
yeoyujie, line 18: Should we extract out tasklist storage and UI? Could consider using Duke.java to control the logic flow instead of exposing storage, UI and tasklist all to parser?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+        this.filePath = filePath;
+        f = new File(this.filePath);
+        tasksList = new ArrayList<>();
+    }
+
+    /**
+     * Reads the tasks from the file and loads them into an <code>ArrayList</code>
+     * @return An <code>ArrayList</code> containing the tasks from the file.
+     */
+    public ArrayList<String> load() {
```
yeoyujie: Could consider renaming this loadFromFile to standardise with appendToFile

yeoyujie (commented):
LGTM, overall this seem v solid.

src/main/java/duke/parser/Parser.java
sunyuting2000: I agree with your criticism here. I will try to reduce the method length when I do have the time to look at this again.

src/main/java/duke/storage/Storage.java
sunyuting2000: Okay thanks for your recommendation!

src/main/java/duke/parser/Parser.java
sunyuting2000: This is true since I am indeed exposing these classes to parser. I will try to address this issue when I have the time.

src/main/java/duke/storage/Storage.java
sunyuting2000: Thank you for the suggestion. I will change it soon.

PR #27, opened 2022-08-19
[Kang Wenhan] iP

onepersonhere:
# Renaissance Duke

> All new original Duke with a twist 

Renaissance Duke is a very **perculiar** chat bot that helps you keep track of your daily task! 🙃

Features:
- Differation of Task
- Search by date
- Search by Keyword
- Load & Save 
- And… more!

Commands:
1. date (search by date)
2. find (search by keyword)
3. deadline (add a dateline)
4. event (add a event)
5. todo (add a todo)
6. delete (delete a task)
7. list (list all the current task)
8. load (loads a task from file)
9. save (saves a task into file)
10. mark (marks the task as done)
11. unmark (unmarks the task as done)
12. exit (exits Duke) 

Download it [here](https://github.com/onepersonhere/ip/releases/tag/A-Jar)!

Future plans:

 - [x]  Managing more tasks
 - [ ] Reminder
 - [ ] Multiple language support
 - [ ] Voiceover
__________

To add more `Commands` to Renaissance Duke, do extend this `abstract` class:
```java
 public abstract class Command {
	protected boolean isExit = false;

    /**
     * Sets the exit flag.
     *
     * @return true if the exit flag is set, false otherwise.
     */
    public boolean isExit() {
        return isExit;
    }

    /**
     * Executes the command.
     *
     * @param ui the user interface
     * @param storageList the storage list
     * @throws DukeException if an error occurs
     */
    public abstract void execute(Ui ui, StorageList storageList) throws DukeException;
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+            case "todo":
+                addTask(new Todo(desc), storageList);
+                break;
+            default:
+            }
+            FileIO.save(storageList, "./default.txt");
+        } catch (DukeException e) {
+            System.out.println(e.getMessage());
+        }
+    }
```
ekweirui: Perhaps this block of code would be better placed in the three addcommand classes, as it might be weird for the other classes to be handled individually but addcommand handled here.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Adds a Task to the StorageList, print out the appropriate String.
+     *
+     * @param task Task to be added to the StorageList
+     * @param storageList StorageList to be added to
+     */
+    private static void addTask(Task task, StorageList storageList) {
+        storageList.add(task);
+        Output.ADD.modifyTask(task, storageList);
+    }
```
ekweirui: Perhaps this piece of code would be better in the storageList class as it is a behaviour of a storageList and not parser itself.

src/main/java/duke/StorageList.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Resets the list of tasks.
+     */
+    public void reset() {
+        list.clear();
+    }
+
+    /**
```
ekweirui, line 101: I like that you implemented several ways to filter the tasks!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,39 @@
...
+ * Add a new Task to the StorageList.
+ */
+public class AddCommand extends Command {
+    /**
+     * Executes the add command.
+     *
+     * @param ui the user interface
+     * @param storageList the storage list
+     * @throws DukeException if the command is invalid
+     */
```
ekweirui, line 18: I feel that it would be tidier and more efficient to put Javadoc comment for just the overridden method in the abstract command class as it can potentially save a lot of your time. (javadoc comments still applies for overriding method)

ekweirui (commented):
Overall, your code looks really good and clean! I admire how you gave more than the specified ways to filter out tasks and print them out. I learnt alot from reading your code and I think it is definitely on the right track. Other than a few small suggestions, I think you did a great job! 😃

src/main/java/duke/gui/MainWindow.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Initialises the main window.
+     */
+    @FXML
+    public void initialize() {
+        scrollPane.vvalueProperty().bind(dialogContainer.heightProperty());
+    }
+
+    /**
+     * Sets the duke object.
```
senwei01, line 42: From coding standard, Javadoc comments require a newline between description and param section. I've only spotted this here, in general you followed the coding standard pretty well

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+    }
+
+    /**
+     * Adds a Task to the StorageList, print out the appropriate String.
+     *
+     * @param task Task to be added to the StorageList
+     * @param storageList StorageList to be added to
+     */
+    private static void addTask(Task task, StorageList storageList) {
```
senwei01: Should this be in StorageList instead? Might want to consider it as it is more of a function of storageList.

src/main/java/duke/Output.java
```diff
@@ -0,0 +1,135 @@
+package duke;
```
senwei01, line 1: I like how you have a class just for the output :)

senwei01 (commented):
Overall I really like the readability of your code. I think you have great OOP as well!

src/main/java/duke/Parser.java
onepersonhere: That is true, however this class is suppose to be able to parse and make sense of the command the user has type, which I why I feel that putting this function in Parser class is more approprite.

src/main/java/duke/Parser.java
onepersonhere: Thanks for the review, this class will be moved to StorageList.

src/main/java/duke/Parser.java
onepersonhere: Thanks for the review, this class will be moved to StorageList.

src/main/java/duke/StorageList.java
onepersonhere: Thanks!

src/main/java/duke/Output.java
onepersonhere: Thanks!

src/main/java/duke/command/AddCommand.java
onepersonhere: Thanks, will do so for future overriden methods!

src/main/java/duke/gui/MainWindow.java
onepersonhere: Thanks for the spot!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Duke is a personal assistant Chat-bot that aims to help users to keep track of various things.
+ */
+public class Duke {
+    private static final String DEFAULT_FILE_NAME = "default.txt";
+    private StorageList storageList;
+    private Ui ui;
+    private ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+    /**
```
Dilysss, line 17: Try to leave lines in between to make the code neater

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+public class Parser {
+    /**
+     * Parses the user command and returns the corresponding command object.
+     *
+     * @param command The user command
+     * @return The corresponding command object
+     * @throws DukeException if the command is invalid
+     */
+    public static Command parse(String command) throws DukeException {
+        switch (command) {
```
Dilysss, line 30: Good that you used switch statements instead of if-else to make the code more readable

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param input User input
+     * @param command First command
+     * @return String between first command and second command (if exist)
+     * @throws DukeException if the command is invalid
+     */
+    public static String findFirstCommand(String input, String command) throws DukeException {
+        int endOfCommand = input.indexOf("/");
+        int beginIndex = input.indexOf(command) + command.length() + 1;
+        if (beginIndex > input.length()) {
```
Dilysss, line 82:
```suggestion
boolean isIncorrectInput = beginIndex > input.length();
```
Similarly for line 81, where the expressions can be simplified

Dilysss (commented):
Well written code!

PR #32, opened 2022-08-19
[Tan Ming Wei] iP

malwaregarry:
Duke is a CLI task list manager. It's
* text-based
* made with ~~Python~~ Java 
### Features
1. Task management
2. *and more...*
### What's planned
- [ ] Graphical UI
- [ ] Code refactoring

Download it [here](https://github.com/malwaregarry/ip/releases) for **FREE**!
 ___
Here's the ```main``` method:

```java
    new Duke(STORAGE_PATH).run();
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,283 @@
...
+import duke.command.FindCommand;
+import duke.command.ListCommand;
+import duke.command.MarkCommand;
+import duke.command.UnknownCommand;
+import duke.command.UnmarkCommand;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
```
fannyjian, line 25: I like how you listed out all your imports to follow the coding standards!

src/main/java/duke/StorageInterface.java
```diff
@@ -0,0 +1,44 @@
...
+
+import duke.task.Task;
+
+/**
+ * Interface for storage functions.
+ */
+public interface StorageInterface {
+
+    /**
+     * Read storage file and return data in file as a list of Tasks.
```
fannyjian, line 13: I think this line should start with "Reads" instead of "Read"!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,283 @@
...
+    private static final String LIST = "list";
+    private static final String TODO = "todo";
+    private static final String DEADLINE = "deadline";
+    private static final String EVENT = "event";
+    private static final String MARK = "mark";
+    private static final String UNMARK = "unmark";
+    private static final String DELETE = "delete";
+    private static final String FIND = "find";
+    private static final String DEADLINE_INDICATOR_PATTERN = "\\s*/by\\s*";
+    private static final String EVENT_INDICATOR_PATTERN = "\\s*/at\\s*";
```
fannyjian: Good naming of constants!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,283 @@
...
+        case UNMARK:
+            command = parseUnmark(getCommandArgument(commandArguments));
+            break;
+        case DELETE:
+            command = parseDelete(getCommandArgument(commandArguments));
+            break;
+        case FIND:
+            command = parseFind(getCommandArgument(commandArguments));
+            break;
+        default:
```
fannyjian: Correct avoidance of indentation for case clauses!

fannyjian (commented):
Overall, your code complies with the coding standards, except for a minor mistake pointed out!

src/main/java/duke/Parser.java
rjkoh: i like how you check the commands using constants, and follow the naming conventions. the indicator patterns are also well done in that they are complicated enough that users will not accidentally input the same pattern due to typos!

src/main/java/duke/StorageStub.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public void updateLine(int lineIndex, String updatedLine) throws DukeException {
+    }
+
+    public List<Task> getTasks() {
+        return this.tasks;
+    }
+}
```
rjkoh: I like the use of stubs for testing, since your Storage class is detailed!

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * {@inheritDoc}
+     * Adds task to TaskList, Storage and displays update message.
+     */
+    @Override
+    public String execute() throws DukeException {
+        Command.taskList.add(this.task);
+        Command.storage.save(this.task);
+        return Ui.getAddTaskMessage(this.task, Command.taskList.size());
+    }
```
rjkoh: I like how the commands are in their own classes and have their own execution implemented in the class itself (instead of being run outside)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,283 @@
...
+     * Parses input into corresponding Command object.
+     *
+     * @param input User input.
+     * @return Command object.
+     * @throws DukeException When error in parsing or executing command.
+     */
+    static Command parse(String input) throws DukeException {
+        Map.Entry<String, Optional<String>> commandString = getCommandAndArguments(input);
+        String commandIndicator = commandString.getKey();
+        Optional<String> commandArguments = commandString.getValue();
```
rjkoh: I like your implementation of parse here as it uses Optional to account for missing or bad inputs. However, it takes some analysis due to the other implemented methods used. 

Perhaps you could consider adding some comments to make it easier to follow as I did have some trouble doing so.

rjkoh (commented):
I enjoyed looking at your code and learnt quite a bit from it! I particularly like how you broke the commands down, and implemented an interface and stub for storage. Your implementation is very detailed and accounts for many potential source of errors!

One suggestion would simply be to add some comments to make your code easier to follow since it is more complicated than most implementations! Well done though!

PR #33, opened 2022-08-19
[Yew Wei Quan] iP

weiquany:
Duke
Duke frees your mind from having to remember things you need to do. It's,
- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://www.youtube.com/watch?v=djV11Xbc914).
2. something
3. something

And it is **FREE**!
 
Features:
- [X] Managing tasks
- [ ] something new (coming soon)
- [ ] something new (coming soon)
___
Here's the `main` method:
```
public class Main {
    }
}
```


src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,71 @@
...
+import java.time.LocalDate;
+
+import duke.DukeException;
+import duke.util.Parser;
+
+/**
+ * Event task.
+ */
+public class Event extends Task {
+    private LocalDate at;
```
yeojunjie: Hello Wei Quan, I think it would be clearer to use a noun as a variable name, as the preposition 'at' could also be used in front of a location.
```suggestion
    private LocalDate startTime;
```

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public boolean contains(String keyword) {
+        return this.description.contains(keyword);
+    }
+
+    /**
+     * Returns a String representation of the {@code Task} in storage format.
+     */
+    public String getStorageFormat() {
+        return this.getStatusBit() + " | " + this.description;
```
yeojunjie: Hi Wei Quan, the method name `getStorageFormat` effectively communicates that this method is meant to encode a `Task` as a `String` for storage. Other authors I reviewed have used names such as `stringify`, which is problematic because it does not self-explain the difference between `stringify` and `toString`.

However, I think it would be clearer to use `String.format()` here. It might be unnecessary to use `getStatusBit()`.

```suggestion
        return String.format("%d | %s", this.isDone ? 1 : 0, this.description);
```

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return numberedTaskList;
+    }
+
+    /**
+     * Returns a {@code TaskList} of all tasks that contains the search keyword in their description.
+     *
+     * @return {@code TaskList} of all tasks containing the search keyword in their description.
+     */
+    public TaskList find(String keyword) {
```
yeojunjie, line 146: Hi Wei Quan, I thought that the naming conventions were well-followed in this method -- the method name `get` is a verb, the temporary variable `tasksMatched` and the parameter `keyword` is self-explanatory.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+
+import duke.DukeException;
+import duke.util.Parser;
+
+
+/**
+ * Task with a deadline.
+ */
+public class Deadline extends Task {
+    private LocalDate by;
```
yeojunjie: Hello Wei Quan, the field `by` could be renamed to the noun `dueDate` to better explain what it represents. Otherwise, `by` could also potentially refer to the person who added this task.

```suggestion
    private LocalDate dueDate;
```

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,71 @@
...
+    public String getStorageFormat() {
+        return "E | " + super.getStorageFormat() + " | " + this.at;
+    }
+
+    /**
+     * Returns a String representation of {@code Event} task in display format.
+     */
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at: " + this.getFormattedDateString() + ")";
```
yeojunjie: Once again, I think using `String.format()` would make it easier to see where whitespace will end up in the display format.

```suggestion
        return String.format("[E]%s (at: %s)", super.toString(), this.getFormattedDateString());
```

yeojunjie (commented):
Hello Wei Quan, I have left some comments regarding code quality guidelines!

src/main/java/duke/task/Event.java
rylzxc: Yup, I agree with the above comment.

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,148 @@
...
+            verifyArgumentsIsNotEmpty(args);
+            return new DeleteCommand(parseTaskIndex(args));
+        case "find":
+            verifyArgumentsIsNotEmpty(args);
+            return new FindCommand(args);
+        case "":
+            return new EmptyCommand();
+        default:
+            return new UnknownCommand();
+        }
```
rylzxc: Hi Wei Quan, I don't think this is a coding standard violation but I was wondering if it is better to declare constants for your respective cases? I suggest this because if you're managing a super large codebase and you somehow have to check for these `inputCommands` , a constant may save you the time taken to find these hard-coded commands. 

Let me know your thoughts! :)

src/test/java/duke/task/TaskListTest.java
```diff
@@ -0,0 +1,121 @@
...
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import duke.DukeException;
+
+public class TaskListTest {
```
rylzxc, line 13: I like your well thought out test cases in your test classes!

src/main/java/duke/task/Event.java
rylzxc: I agree with this and the similar comment made on the deadline class

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+import java.time.LocalDate;
+
+import duke.DukeException;
+import duke.util.Parser;
+
+
+/**
+ * Task with a deadline.
+ */
+public class Deadline extends Task {
```
rylzxc, line 12: Hi, again, not a coding standard violation but I was wondering if adding Task behind the respective classes for tasks will be better? I feel it allows readers to easily navigate to the type of task they want to see. Also, Event could be misleading as it can represent things like button onClick events etc

```suggestion
public class DeadlineTask extends Task {
```

rylzxc (commented):
Overall, clean code, simple to understand.

PR #37, opened 2022-08-19
[Clarence Chew] iP

clarence-chew:
### Anthea
---
Ever felt you wanted to track your life better? Feeling confused about what you want to do?

> _Where am I?_ - Anthea

Not you, although you are the star of the show here, I mean **you**, dear reader.

Anthea is a chatbot that helps you keep track of your tasks.

> _But, I haven't done that before!_ - Anthea

Don't worry, you can access their filesystem if you need to record things for them.

> _What if I mess up?_ - Anthea

I'll help. For the people you are helping to track their tasks, they would also help you notice bugs.

> _Bugs? Oh no!_ - Anthea

Such is how it goes when it's code made by humans. You know that, don't you? You are made with Java, which looks something like this:

```java
    System.out.println("Hello World!");
}
```

> _Hold up, I remember that._
>
> _In fact, I remember a lot of other things too, like the `Todo`, `Deadline` and `Event` classes._
> * _The `ToDo` class helps you track a task._
> * _The `Deadline` class helps you track a deadline, so you know when to rush._
> * _The `Event` class helps you track events you might need to attend._

Yeah, that is in the code. There are also various prompts that you could respond to as well, such as
1. `list` to list down all the tasks
1. `find` to search for tasks containing a keyword
1. `delete` to remove tasks
1. `mark` and `unmark` to keep track of tasks.

Something like this:
- [x] A task
- [x] Another task
- [ ] Visit [this place](https://nus-cs2103-ay2223s1.github.io/website/index.html) to learn software engineering

I think you are getting it. 😄

> _Yes, I think I am ready. 🦌_ - Anthea


src/main/java/duke/CommandMatcher.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class CommandMatcher {
+    private Predicate<String> shouldRunAction;
+    private Consumer<String> action;
+
+    /**
+     * Creates an object that handles checking and executing a command.
+     * @param shouldRunAction Predicate to check if the command should be run.
+     * @param action Action to run.
+     */
```
cfyjoanna: For JavaDocs, there should be an empty line between description and parameter section. No issues otherwise!

cfyjoanna (commented):
Apart from the JavaDocs, looks like you followed the coding standard really closely. Nice work!

src/main/java/duke/CommandMatcher.java
clarence-chew: Thanks - made those changes!

src/main/java/duke/CommandMatcher.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/**
+ * This class serves as a way to abstract the idea of making a command
+ * as a matching process and an action.
+ */
+public class CommandMatcher {
+    private Predicate<String> shouldRunAction;
+    private Consumer<String> action;
```
marcuslowhuiyu, line 12: I like how you use the consumer functional class in your implementation, seems to be a rather unique way to do it

src/main/java/duke/ParsedDateTime.java
```diff
@@ -0,0 +1,52 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.Optional;
+
+/**
+ * Helper class to parse date/time strings
+ */
+public class ParsedDateTime {
+    private static final DateTimeFormatter[] formatters = {
```
marcuslowhuiyu, line 12: I like how you had a way to test for various date time formats

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,98 @@
...
+            Scanner sc = new Scanner(f);
+            while (sc.hasNext()) {
+                ArrayList<String> curLine = new ArrayList<>();
+                for (String str : sc.nextLine().split(",")) {
+                    curLine.add(convertFromBase64(str));
+                }
+                lines.add(curLine.toArray(new String[]{}));
+            }
+            return (String[][]) lines.toArray(new String[][]{});
+        } catch (FileNotFoundException ex) {
```
myangat0343: Everything is good!!! Just one small suggestion that maybe you can set up some exceptions classes, (e.g. DukeException extends Exception) and put all exception class into one package : )

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,66 @@
+package duke.task;
+
+/**
+ * A task stores the state and description of a task.
```
myangat0343: You can remove "of a task" : )

myangat0343 (commented):
Some suggestions. Hope they could help you : )

src/main/java/duke/task/Task.java
clarence-chew: Your suggestion sounds better.

src/main/java/duke/Storage.java
clarence-chew: The existing exceptions I find a pretty useful, so no need to duplicate them. Perhaps a Response class to ensure that we print only once can be useful. Perhaps having it throwable like an exception can help too, so that once it is decided it can skip past the happy path to land in a catch block.

clarence-chew: Done as stated. It was quite a hassle.

Do I regret it? I guess not.

PR #43, opened 2022-08-20
[Aaron Li] ip

aaronliwh:
 - text-based
 - easy to learn
 - ~~FAST~~ SUPER FAST to use
 
 All you need to do is,
2. double-click it.
3. add your tasks.

And it is FREE!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)  
 
#

```java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+public class Duke {
+    static Ui ui = new Ui();
+    static Storage storage = new Storage();
+    static TaskList taskList = new TaskList(storage.readFile("duke.txt"));
+
+
+    public String getResponse(String input) {
+        try {
+            Command c = Parser.parse(input);
```
RussellDash332: Since you have a `parseTask` method somewhere in this repo, do you think it's a good idea to rename this method to `parseInput`?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+                LocalDate dateDeadline = getDate(stringDateDeadline);
+                return new AddCommand("deadline",nameDeadline,dateDeadline);
+            case "event" :
+                int indexOfDateEvent = findDate(input);
+                String nameEvent = String.join(" ", Arrays.copyOfRange(input, 1, indexOfDateEvent));
+                String stringDateEvent = input[indexOfDateEvent + 1];
+                LocalDate dateEvent = getDate(stringDateEvent);
+                return new AddCommand("event",nameEvent,dateEvent);
+            case "mark" :
+                int indexMark = getNumber(input);
```
RussellDash332: (Analogously for the cases below)
Perhaps `markIndex` would be a more sensible naming since it really sounds like a noun?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,98 @@
...
+            returnString += (index + ". " + taskList.get(i));
+        }
+        return returnString;
+    }
+
+    /**
+     * Returns the number of tasks in the list
+     * @return the size of the List of Tasks
+     */
+    public int size(){
```
RussellDash332: What do you think of `getSize` instead of `size`? It might be clearer to even more unfamiliar readers

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,79 @@
...
+     * @return The String representation of the user's input
+     */
+
+
+    /**
+     * Prints the appropriate message for adding a Task
+     * @param task The task added
+     * @param size The current number of tasks
+     */
+    public String showAddCommand(Task task,int size) {
```
RussellDash332: Both of these methods are returning a String, but only the latter is a verb. Should you rename `goodbyeMessage` to a verb as well?

RussellDash332 (commented):
LGTM! Very little nits to deal with since my suggestions so far are somewhat not an urgent one but probably good to consider 👍🏻

Loving the GUI resources as well 💪🏻

src/main/java/duke/Parser.java
darrentde: I agree with @RussellDash332 suggestion that `markIndex` may be a better name

darrentde: I agree with @RussellDash332 suggestion that `markIndex` may be a better name

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,98 @@
+package duke;
```
darrentde, line 1: Perhaps you could start to create more specified packages for your files since you have already done it for command?

darrentde (commented):
LGTM. In general your code meets the naming conventions. 🤩

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+     * Parses the input and returns the appropriate command to be executed
+     * @param inputString the inputted string from the user
+     * @return the Command to the executed from the input
+     */
+    public static Command parse(String inputString) throws DukeException {
+        String[] input = inputString.split(" ");
+        switch (input[0]) {
+            case "bye":
+                return new EndCommand();
+            case "list":
```
RachelChua: Small issue but based on Java coding standard guidelines, I think there should not be indentations for case clauses. 😃

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return The Task stored
+     */
+    private Task stringToTask(String data) throws DukeException {
+        if(data.length() == 0){
+            return null;
+        }
+        char type = (data.charAt(0));
+        boolean done = (data.charAt(1)) == ('Y');
+        Task task = null;
+        if(type == 'T'){
```
RachelChua: Seems to have a missing space between if and the round opening bracket, as well as the closing round bracket and the curly brackets.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,98 @@
...
+    public List<Task> getTaskList(){
+        return taskList;
+    }
+
+    /**
+     * Prints the list of Tasks
+     */
+    public String printTaskList(){
+        String returnString = "";
+        for(int i = 0; i<taskList.size(); i++){
```
RachelChua: Should there be spaces between the brackets?

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,24 @@
...
+
+import duke.DukeException;
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+
+/**
+ * Encapsulates a command.
+ */
+public abstract class Command {
```
RachelChua, line 11: I like your abstraction for the command class!

RachelChua (approved):
Just a few minor coding standards issues but overall, I think that your OOP and abstraction is done well! Good job! 👍

PR #55, opened 2022-08-20
[Donovan Lee] iP

24Donovan24:
* text-based
* easy to learn
* ~FAST~ *SUPER* FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.
4. let it manage your tasks for you :stuck_out_tongue_winking_eye:

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
---
```java
public class Main {
    }
}
```


src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Displays the todo task added with its corresponding index.
+     *
+     * @param todo The todo task specified by user.
+     * @param size The number of tasks in the list.
+     */
+    public void printTodo(Task todo, int size) {
+        System.out.printf(LINE_BREAK
+                + "\n\tGot it. I've added this task:\n\t%s\n\tNow you have %d tasks in the list.\n"
+                        + LINE_BREAK + "\n", todo, size);
```
ElijahS67: Nice use of line breaks to improve readability!

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,59 @@
...
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+
+    /**
+     * Creates a string to be saved in the file.
+     *
+     * @return String to be displayed in the file.
+     */
+    public abstract String savedString();
```
ElijahS67: Perhaps toSavedString() is a better method name?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,97 @@
...
+     * Parses the various inputs inputted by user.
+     */
+    public void start() {
+        Scanner input = new Scanner(System.in);
+        String text = input.next();
+        while (!text.equals("bye")) {
+            try {
+                switch (text) {
+                //Handle case when task aTodo
+                case "todo" :
```
ElijahS67: Perhaps there should not be a spacing between "todo" and the colon?

ElijahS67 (commented):
Looks good! Almost LGTM!

src/main/java/duke/Task.java
24Donovan24: Thanks for the review! I will implement the following suggestion!

src/main/java/duke/Parser.java
24Donovan24: Thanks for the review! I will make that change!

PR #56, opened 2022-08-20
[Tina Agarwal] iP

Thing1Thing2:
# iP Pull Request
###### By Tina Agarwal/ Thing1Thing2
**In response to** :
> Week 4 [Fri, Aug 26th] - Project
iP: Use GFMD in the PR description

**How to use?**
- Clone repo `git clone`
- Run build

**Changes made:**
1. Change name to "Falcon"
2. Add Gradle
3. Rectify Issues found using Checkstyle
4. Add JUnit tests

:page_facing_up: **Duke class overview**:
```java
public class Duke {
    private Storage storage;
    private TaskList tasks;
    private UI ui;

    public Duke(String filePath) { ... }
    public void run() {...}
    public static void main(String[] args) {...}
}

```

:alarm_clock: **Increments pending as of [Week 4](https://nus-cs2103-ay2223s1.github.io/website/schedule/week4/project.html)**:
- [ ] Level-10
- [ ]  A-Checkstyle
- [ ]  A-Varargs
- [ ] A-Gradle
- [ ] A-JAR


src/main/java/Duke.java
```diff
@@ -1,10 +1,55 @@
...
 public class Duke {
+
+    private Storage storage;
+    private TaskList tasks;
+    private UI ui;
+
+    /**
+     * Constructor.
+     * @param filePath Path where disk file should be created.
+     */
```
JonathanWiguna: I think there should be a line between the description and the parameter section of your JavaDocs.

src/main/java/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import utility.Parser;
+
+/**
+ * Command that handles adding Deadline to TaskList and Storage.
+ */
+public class AddDeadlineCommand extends Command {
+
+    /**
+     * Runs when deadline is to be added.
+     * @param taskList TaskList to append Deadline to.
```
JonathanWiguna: Same here

src/main/java/Duke.java
```diff
@@ -1,10 +1,55 @@
+import command.Command;
```
JonathanWiguna: Shouldn't Duke be in a package as well?

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Class representing disk storage used in Duke
+ * to backup task list.
+ * Requires valid file parh to be set at
+ * each run of Duke.
+ */
+public class Storage {
+    private Path path;
+    private final StorageReader storageReader;
+    private final StorageWriter storageWriter;
```
JonathanWiguna, line 20: Since these are constants, I think they should be in all uppercase with underscore to separate the words

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,86 @@
...
+        return storageWriter.appendLine(s);
+    }
+
+    public boolean deleteLine(int index) {
+        return storageWriter.deleteLine(index);
+    }
+
+    public boolean changeLine(int index, String newString) {
+        return storageWriter.changeLine(index, newString);
+    }
```
JonathanWiguna: Boolean methods should sound like booleans right?

JonathanWiguna (commented):
Looks good! Just some small changes regarding coding practice requested.

src/main/java/storage/Storage.java
Thing1Thing2: Ah yes, thank you for mentioning that :D

src/main/java/storage/Storage.java
Thing1Thing2: True, I got a little confused since changeLine causes change as well as returning true if the change was successful.

PR #57, opened 2022-08-20
[Chu Thi Thanh] iP

lilythchu:
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com)
2. double-click it
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

If you Java programmer, you can use it to pratice Java too. Here's the <span style="color:red">`main`</span> method:
```
public class Main {
  }
}
```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,88 @@
...
+public class Storage {
+    private final String filePath;
+    private final String dirPath;
+
+    /**
+     * Constructor for Storage.
+     */
+    public Storage() {
+        this.filePath = "data" + File.separator + "lilychat.txt";
+        this.dirPath = "data";
```
PangKuangWei, line 28: The "Java coding standard (all)" page says that we should use the `this` keyword only when a field is shadowed by a method or constructor parameter to reduce unnecessary noise.

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,99 @@
+package duke;
+
+import java.util.Scanner;
+
+import duke.task.Task;
+
+/**
+ * Represents the UI of Duke.
+ */
+public class UI {
```
PangKuangWei, line 8: I am unsure whether this is a problem but I would recommend changing this to "Ui" because `UI` could be mixed up with constants. The example given on the Week 3 task uses "Ui" so it should be safer to use it.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,205 @@
...
+        } else if (request.matches("(?i)^(find)(.*)")) {
+            return RequestType.FIND;
+        } else if (request.matches("(?i)^(deadline)(.*)")) {
+            return RequestType.DEADLINE;
+        } else if (request.matches("(?i)^(event)(.*)")) {
+            return RequestType.EVENT;
+        } else {
+            throw new InvalidInputException();
+        }
+    }
```
PangKuangWei: I like your usage of the matches method with regex. If the functionalities expand you could consider using a switch for the single word commands that you use the equals method for and the remaining if else stack in the default case.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,49 @@
...
+        this.taskList = new TaskList(storage.initialise());
+    }
+
+    /**
+     * Runs the duke program.
+     */
+    public void run() {
+        ui.greet();
+        boolean isExit = false;
+        while (!isExit) {
```
PangKuangWei, line 30: Perhaps changing `isExit` to "canExit" would be more suitable?

PangKuangWei (commented):
Your code looks good! I feel that your code was easy to read as it was very organised. It was very hard to find things to comment about.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+import duke.Storage;
+import duke.TaskList;
+import duke.UI;
+
+/**
+ * Representation of the find command.
+ */
+public class FindCommand extends Command {
+    private final String keyword;
```
DarrenCsAcc, line 11: Good use of access modifiers

PR #59, opened 2022-08-20
[Justin Siaw] ip

JustinS00:
# Duke

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. install java
```java
	System.out.println("Hello World");
}
```
2. run Duke
3. type commands
4. see the magic 🪄 

## Features
- [X] Managing tasks
- [X] `delete` tasks
- [X] `find` tasks


src/main/java/duke/command/DukeCommandType.java
```diff
@@ -0,0 +1,17 @@
+package duke.command;
+
+/**
+ * Enumeration of the type of commands available in Duke
+ */
+public enum DukeCommandType {
```
comicalromance, line 6: Great use of enum in a separate class to clarify what each method is about!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Returns an array list of tasks stored from previous runs
+     *
+     * @return List of Tasks
+     * */
+    public static ArrayList<Task> loadData() {
+        try {
+            FileInputStream fin = new FileInputStream(FILE_NAME);
+            ObjectInputStream ois = new ObjectInputStream(fin);
+            @SuppressWarnings("unchecked") ArrayList<Task> tasks = (ArrayList<Task>) ois.readObject();
```
comicalromance, line 30: It'd be great if you could also close the FileInputStream after you're done with it (a 'try with resources' statement would be very convenient for your usecase!)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,58 @@
...
+        case "todo":
+            return DukeCommandType.TODO;
+        case "deadline":
+            return DukeCommandType.DEADLINE;
+        case "event":
+            return DukeCommandType.EVENT;
+        case "find":
+            return DukeCommandType.FIND;
+        default:
+            return DukeCommandType.UNKNOWN;
```
comicalromance, line 56: I really like how you handled the default case instead of just leaving it to fail / cause an error!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,81 @@
...
+        run();
+    }
+
+    private static void run() {
+        while (dukeUi.hasNextLine()) {
+            Command command = Parser.getCommand(dukeUi.getNextLine());
+            DukeCommandType commandType = command.getCommandType();
+            String args = command.getArgs();
+            switch (commandType) {
+            case EXIT: {
```
comicalromance: Might be personal preference but the braces on switch statements are optional and so you could technically remove them if you want to! Same goes for the other switch statements as well.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,59 @@
...
+            return LocalDateTime.parse(deadline).format(DateTimeFormatter.ofPattern("dd MMM yyyy HH:mm"));
+        } else if (Pattern.compile("([0-9]{4})[-]([1][0-2]|0[1-9])[-](3[01]|[12][0-9]|0[1-9])")
+                .matcher(deadline).matches()) { //yyyy-mm-dd
+            return LocalDate.parse(deadline).format(DateTimeFormatter.ofPattern("dd MMM yyyy"));
+        } else if (Pattern.compile("(3[01]|[12][0-9]|0[1-9])[-]([1][0-2]|0[1-9])[-][0-9]{4}")
+                .matcher(deadline).matches()) { //dd-mm-yyyy
+            String dd = deadline.substring(0, 2);
+            deadline.replace(0, 2, deadline.substring(6, 10)).replace(8, 12, dd);
+            return LocalDate.parse(deadline).format(DateTimeFormatter.ofPattern("dd MMM yyyy"));
+        }
```
comicalromance, line 48: This section may be too complicated (especially the if statements). Would it help if you perhaps initialize the patterns as individual variables first (with names to indicate the pattern used) and then use it in the conditions?

comicalromance (commented):
LGTM! The structure and flow of the code seems very functional and I feel the OOP has been definitely executed well.

src/main/java/duke/command/DukeCommandType.java
albertZhangTJ, line 6: I really like how you used enum to differentiate all kinds of commands.

albertZhangTJ (commented):
I really like your work! I think its quite sophisticatedly coded and it is really hard for me to find style errors!
 One suggestion I can come up with is you might want to add `@author` tags to your class-level JavaDoc comments

PR #83, opened 2022-08-20
[Koh Chen Huang, Samuel] iP

Samsation:
# dukeBot Ver. Pingu
>"A computer once beat me at chess, but it was no match for me at kick boxing." – Emo Philips
([source](https://www.brainyquote.com/quotes/emo_philips_128993))

dukeBot Ver. Pingu frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn

All you need to do is, 
1. download it from [here](https://github.com/Samsation/ip)
2. double-click it
3. add your tasks

And it is **FREE**!

## Features
- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders (coming soon)

```
        Application.launch(Main.class, args);
    }

### CODE REVIEW

File: src/main/java/duke/DialogBox.java [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        dialog.setText(text);
+        displayPicture.setImage(img);
+    }
+
+    /**
+     * A method that flips the dialog box such that the ImageView is on the left and text on the right.
```
lkwlkww: Perhaps you could change your method header to start with 'Flips'? According to the provided [coding standards](https://se-education.org/guides/conventions/java/intermediate.html), method headers should start with words like 'Returns', 'Sends', 'Adds' etc

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,106 @@
...
+     * A method that creates a new Command according to the user input.
+     *
+     * @param inputCommand Input command from the user.
+     * @return A specific-typed Command corresponding to the user input.
+     * @throws DukeException If the user input is unrecognised.
+     */
+    public static Command parse(String inputCommand) throws DukeException {
+        try {
+            String[] splitInputCommand = inputCommand.trim().split(" ", 2);
+            CommandTag ct = CommandTag.valueOf(splitInputCommand[0].toUpperCase());
```
lkwlkww: According to the coding standards:
> scratch variables used for temporary storage or indices can be kept short

I'm not sure if `ct` here fits the criteria, or if it should be changed to something more meaningful like `commandTag`. What do you think?

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    private void flip() {
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
+        Collections.reverse(tmp);
+        getChildren().setAll(tmp);
+        setAlignment(Pos.TOP_LEFT);
+    }
+
+    /**
+     * A method that retrieves the user's dialog.
```
lkwlkww: Perhaps you could consider not including method headers for getters in future, if you would like to save some effort:
> You MUST write header comments for all classes, public methods. But they can be omitted for the following cases:
i. Getters/setters

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,52 @@
...
+    private Ui ui;
+    private boolean isTerminated = false;
+
+    /**
+     * A constructor for Duke.
+     *
+     * @param stage The current Stage of the Application.
+     */
+    public Duke(Stage stage) {
+        this.ui = new Ui(stage);
```
lkwlkww: I think you could consider leaving the `this` keyword out, according to the following coding standard rule:
> Use the this keyword only when a field is shadowed by a method or constructor parameter.

lkwlkww (commented):
Very well-adhered to the coding standards! It was hard to find deviations.

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,84 @@
...
+
+        dialog.setText(text);
+        displayPicture.setImage(img);
+    }
+
+    /**
+     * A method that flips the dialog box such that the ImageView is on the left and text on the right.
+     */
+    private void flip() {
+        ObservableList<Node> tmp = FXCollections.observableArrayList(this.getChildren());
```
ThomasHoooo: Maybe you can set a more meaningful name instead of just tmp?

ThomasHoooo (commented):
Inconsistent styling of class description.

src/main/java/duke/DialogBox.java
Samsation: Fixed under A-CodeQuality. Thank you!

src/main/java/duke/DialogBox.java
Samsation: Fixed under A-CodeQuality. Thank you!

PR #87, opened 2022-08-20
[Elijah] iP

ElijahS67:
## Duke

> This is a quote.

Duke is your friendly chatbot that helps you manage your tasks. It's

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is

1. download Duke [here](https://github.com/ElijahS67/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

and let it manage your tasks for you! 👍 
And it's **FREE**!

Features:
- [ ] Managing tasks
- [ ] Reminders (coming soon)

```java
public class Main {
        Application.launch(Duke.class, args);
    }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+                if (tArray.length > 2) {
+                    throw new DukeException("Date and time has too many spaces!");
+                } else if (tArray.length == 2) {
+                    time = tArray[1];
+                }
+
+                date = tArray[0];
+
+                //matches the yyyy-MMM-d format
+                if (!date.trim().matches("[0-9]{4}-"
```
teoyuqi: The Regex check here seems to be doing the same thing as the `try-catch` block at line 101. Might it be possible to combine the two?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            if (taskname.equals("")) {
+                throw new DukeException("Name of task cannot be empty!");
+            }
+
+            if (str.equals("todo")) {
+                this.taskList.addToDo(taskname);
+            } else {
+                if (date.equals("")) {
```
teoyuqi: Could we perhaps restructure the `if-else` blocks to avoid using a triply-nested `if-else` block here? 

Alternatively, you may consider using a `switch-case` block with a `case` for each of the command. In that case, the necessary `if-else` checks for each command can be placed in its own `case` block without [deep nesting](https://nus-cs2103-ay2223s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,53 @@
...
+
+        String[] strArray = sc.nextLine().split(" ");
+
+        while (!strArray[0].equals("bye")) {
+            this.parser.parseCommand(strArray);
+
+            strArray = sc.nextLine().split(" ");
+        }
+
+        System.out.println("Bye. Hope to see you again soon!");
```
teoyuqi: Perhaps we could abstract the `println()` statements of all classes to `Ui.java`, since `Ui.java` should be the one handling the user interface side of things?

teoyuqi (commented):
Looks good to me, other than a few minor changes here and there to improve readability.

I love how your code is kept clean and simple, keep it up! 👍🏻

src/main/java/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,3 @@
+Manifest-Version: 1.0
+Main-Class: duke.Duke
```
Tan-Jia-Rong, line 2: Good that you use lower caps for package!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Represents a Deadline object that is also a Task.
+ * Inherits from the Task class.
+ *
+ */
+public class Deadline extends Task {
+
+    private LocalDate time;
+    private final char type = 'D';
```
Tan-Jia-Rong: Nice! I like how you specify the type.
However, I believe it could be static final since the type never changes for the class Deadline
Apart from that do use caps when define final constants
```suggestion
    private static final char TYPE_DEADLINE = 'D';
```

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Represents an Event object that is also a Task.
+ * Inherits from the Task class.
+ *
+ */
+public class Event extends Task {
+
+    private LocalDate time;
+    private final char type = 'E';
```
Tan-Jia-Rong: Same comment as above!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+        Scanner sc = new Scanner(savedFile);
+        ArrayList<Task> arrayList = new ArrayList<>();
+
+        while (sc.hasNextLine()) { //type#boolean#taskname#time
+
+            String[] array = sc.nextLine().split("#");
+            boolean isToDo = (array.length == 3 && array[0].equals("T"));
+            boolean isDeadlineOrEvent = (array.length == 4 && (array[0].equals("D") || array[0].equals("E")));
+            boolean hasCorrectMark = (array[1].equals("1") || array[1].equals("0"));
+            boolean isValidTask = isToDo || isDeadlineOrEvent;
```
Tan-Jia-Rong, line 82: good use of prefix to specify boolean!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private static File savedFile = null;
+
+    /** Constructor for a Storage object */
+    protected Storage() {
+
+    }
+
+    /**
+     * Is called upon starting Duke. Attempts to find a save file.
```
Tan-Jia-Rong: Should start with a verb
is called upon starting duke can be left at description
```suggestion
     * Attempts to find a save file.
```

Tan-Jia-Rong (changes requested):
LGTM! Aside for some minute details, really like how you structure ur classes 😄

src/main/java/duke/Parser.java
ElijahS67: Thanks for the review, will update it!

src/main/java/duke/Deadline.java
ElijahS67: Thanks for the review, will update it!

src/main/java/duke/Storage.java
turretDive, line 82: I like how you used a question for all your variable names for boolean! It is a very common mistakes that I would sometimes miss out myself!

turretDive (commented):
LGTM! Pretty much everything about coding standards have already been said by other reviewers.

PR #91, opened 2022-08-20
[Tan Chee Heng] iP

cheeheng:
## Jude

> Hey Jude, don't make it bad.
> Take a sad song and make it better. - lyrics of pop song hit [_Hey Jude_](https://www.youtube.com/watch?v=A_MjCqQoLLA), The Beatles (1968)

Fast forward to 2022, and Jude the chatbot is born. Jude is a
- Text-based command-line interface
- Task tracker which helps track events and deadlines

To use Jude the chatbot, do the following:
1. Download the file Jude-0.1.jar [here](https://github.com/cheeheng/ip/releases)
1. Type the command `java -jar Jude-0.1.jar` on the ~different~ **same** directory as Jude-0.1.jar
1. Add tasks
1. Jude the chatbot will help you manage the tasks 😊

Here are some features of Jude the chatbot:
- [x] Add and delete tasks
   - [x] All tasks have a description
   - [x] Todo tasks
   - [x] Deadline tasks have a deadline which needs to be a proper date
   - [x] Event tasks have a start and end time which is currently condensed under 1 parameter /at 
- [ ] Sort deadline tasks by deadline (not yet implemented)
- [ ] Event tasks have a separate start and end time (not yet implemented)

This is the ```main()``` method which was used in an earlier version of the chatbot:
```java
/**
 * Runs the task tracker chatbot.
 *
 * @param args Not used for now.
 */
public static void main(String[] args) throws IOException {
    new Jude("data/tasks.txt").run();
}
```


src/main/java/jude/Main.java
```diff
@@ -0,0 +1,48 @@
...
+
+//@@author cheeheng-reused
+//Reused from https://se-education.org/guides/tutorials/javaFxPart4.html with minor
+//modifications.
+
+/**
+ * A GUI for Jude the chatbot using FXML.
+ */
+public class Main extends Application {
+    private final Jude jude = new Jude("data/tasks.txt");
```
Josephaedan: For easier modification, perhaps you could consider refactoring 'data/tasks.txt' into a constant TASKS_LIST or something outside the code just for easier readability/editing in the future!

src/main/java/jude/Parser.java
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * @param command The command which will be provided to the parser.
+     * @return true if the command terminates the program, i.e. bye command, false otherwise.
+     * @throws IOException When system I/O fails.
+     */
+    public String parse(String command) throws IOException {
+        command = command.trim(); // to remove unintended trailing spaces
+        String[] tokens = command.split(" ", 2);
+        String response = "";
+        try {
```
Josephaedan, line 140: You wrote this function to parse and execute the command provided. However, as a result, it is really long and difficult to read the different cases. Perhaps you could consider refactoring out the different execution codes to separate functions so that it is much cleaner and readable.

src/main/java/jude/task/Task.java
```diff
@@ -0,0 +1,110 @@
...
+    //Reused from https://nus-cs2103-ay2223s1.github.io/website/schedule/week2/project.html
+    //with minor modifications
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    /**
+     * Sets the status of the {@code Task} to 'done'.
+     */
+    public void markAsDone() {
```
Josephaedan, line 56: It seems like markAsDone(), markAsUndone() and setIsDone(isDone) all serve the purpose of setting this.isDone property. I'm not really sure about the value of setIsDone() being executed in other parts of your codebase as this is bad practice; Ideally it should only be called within the class itself. Perhaps you could consider just removing setIsDone() and using markAsDone() and markAsUndone() for the rest of the codebase. That way you can do proper error handling within the functions themselves too.

Josephaedan (commented):
Overall, I think your code looks good! My only concern is for people reviewing your code in the future, understanding the parsing and executing of the command may be a little too tedious and difficult so you could consider abstracting them into separate functions.

cheeheng:
Thank you for your input. Yes, I agree that the parsing code is a bit too long and I would probably have to refactor this part into several functions.

src/main/java/jude/Parser.java
cheeheng: I agree with this. In addition, line 117 contains a redundant comment which I should remove.

src/main/java/jude/Storage.java
```diff
@@ -0,0 +1,113 @@
...
+
+            Task task;
+            if (taskType.equals("T")) {
+                task = new Todo(taskName, isDone);
+            } else if (taskType.equals("D")) {
+                String deadline = sc.nextLine();
+                task = new Deadline(taskName, isDone, deadline);
+            } else if (taskType.equals("E")) {
+                String eventTime = sc.nextLine();
+                task = new Event(taskName, isDone, eventTime);
```
yuxuanleong: This can probably be made into a method on itself and the code will look cleaner 😄

yuxuanleong (commented):
You are generally doing a really good job, just some tidying up jobs for longer methods!

PR #92, opened 2022-08-20
[Jicson Toh] iP

jicsontoh:
# DukePro

* text-based
* easy to learn
* ~~FAST~~ _SUPER_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/nus-cs2103-AY2223S1/ip/pull/92).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

If you Java programmer, you can use it to practice Java too. Here's the `main `method:
```Java
public class Main {
    }
}
```


src/main/java/duke/command/UnmarkCommand.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param action user input action.
+     */
+    public UnmarkCommand(String action) {
+        this.action = action;
+    }
+
+    /**
+     * Executes the command input.
+     * @param tasks   amends task list if any.
+     * @param ui      ui to output feedback.
```
JovonLim, line 26:
```suggestion
     * Executes the command input.
     *
     * @param tasks   amends task list if any.
     * @param ui      ui to output feedback.
```
I think there should be an empty line between the description and parameter lines for the coding standard. I noticed that only some docs have that so it would be good to rectify the rest.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Objects;
+
+import duke.util.DukeException;
+
+/**
+ * Task Class to store information of task.
+ * @author Jicson Toh
+ */
+public class Task {
+    protected String action;
```
JovonLim, line 12:
```suggestion
    protected String description;
```
Perhaps this would be a better name rather than action?

src/main/java/duke/data/TaskList.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public TaskList() {
+        this.list = new ArrayList<>();
+    }
+
+    public ArrayList<Task> getList() {
+        return this.list;
+    }
+
+    public String getTaskString(int index) {
```
JovonLim, line 68:
```suggestion
    public String getTaskDescription(int index) {
```
Rather than String, it might be better to refer it as the task's description

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+ * @author Jicson Toh
+ */
+public class Parser {
+
+    /**
+     * Checks the input command.
+     * @param action user input.
+     * @return returns the string action.
+     */
+    public Command parseCommand(String action) throws DukeException {
```
JovonLim, line 35: I like that it systematically checks for the single word commands then for the other commands 👍

src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,126 @@
...
+        System.out.println("OOPS!!! I'm sorry, but I don't know what that means :-(\n"
+                + "--------------------------");
+    }
+
+    /**
+     * Prints the removed task from the list and updates the list count.
+     * @param removedTask task to be removed.
+     * @param size size of list after task is remove.
+     */
+    public void removeTask(String removedTask, int size) {
```
JovonLim:
```suggestion
    public void showRemovedTask(String removedTask, int size) {
```
Since this method is used by the Ui object, perhaps it would make more sense by naming it to indicate that its printing out a message rather than actually removing a task? Similarly for the below task related methods

JovonLim (commented):
Overall good job on the iP 😄. Just take note of the javadocs for the coding standard as well as a few possible improvements to the naming of your methods and fields.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,71 @@
...
+import duke.task.TodoTask;
+import duke.util.DukeException;
+import duke.util.Ui;
+
+/**
+ * Executes the command to add a task to the list.
+ * @author Jicson Toh
+ */
+public class AddCommand extends Command {
+    private final String action;
```
kangqiao322, line 16: Should constant names be in lower case? Some other constant names have the same issue too.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,71 @@
...
+                int j = action.indexOf('/');
+                String by = action.substring(j + 3).strip();
+                tasks.addTask(new DeadlineTask(action.substring(8, j).strip(), by));
+                break;
+            default:
+                throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+            ui.addedTask(tasks.getTaskString(tasks.getSize() - 1), tasks.getSize());
+        } catch (Exception e) {
+            throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
```
kangqiao322, line 58: I like how you used switch cases for numerous different conditions instead of if else. The indentation is correct as well. Great job!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,71 @@
...
+    public boolean getDone() {
+        return isDone;
+    }
+
+    /**
+     * Checks if the task contains the keyword.
+     * @param keyword input keyword.
+     * @return true if contains the keyword.
+     */
+    public boolean matchKeyword(String keyword) {
```
kangqiao322, line 48: Perhaps try to use prefix such as is, has, was etc for boolean method names.

kangqiao322 (commented):
Overall, a job well done! Just need a few changes to the naming of some variables and methods along with more JavaDocs for some public methods.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,63 @@
...
+import duke.task.Task;
+import duke.util.DukeException;
+import duke.util.Ui;
+
+/**
+ * Executes the command to find a task
+ * @author Jicson Toh
+ */
+public class FindCommand extends Command {
+    private final String action;
```
kangqiao322, line 16: I think this variable name should be in upper case as well?

src/main/java/duke/command/UnmarkCommand.java
jicsontoh: Will take note thanks!

src/main/java/duke/task/Task.java
jicsontoh: agreed, will take note thanks!

src/main/java/duke/data/TaskList.java
jicsontoh: agreed, will take note thanks!

src/main/java/duke/task/Task.java
jicsontoh: Will take note thanks!

src/main/java/duke/command/AddCommand.java
jicsontoh: Thank you!

src/main/java/duke/util/Parser.java
jicsontoh: Thanks!

src/main/java/duke/util/Ui.java
jicsontoh: agreed, will take note thanks!

src/main/java/duke/command/AddCommand.java
jicsontoh: hmm its not a constant but a variable to be instantiated when creating the AddCommand Object

src/main/java/duke/command/FindCommand.java
jicsontoh: same for this!

PR #93, opened 2022-08-20
[Joanna] iP

cfyjoanna:
# Duke

Let Duke help you to store your plans so you never have to. Duke is
- text-based
- intuitive for everyone
- *completely **free***!

All you have to do is:
1. download it from [here](https://github.com/cfyjoanna/ip)
2. open the app
3. add your todos, deadlines and events
4. let Duke take care of the rest! :hugs:

## Features
- [x] manage tasks
- [ ] set reminders (coming soon)

## For programmers
You can change Duke to behave however you like too. Here's the `main` method:

```java
public class Duke {
    // ...
        new Duke().run();
    }
}
```


src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public String showWelcome() {
+        return "Hello! I'm Duke.\n"
+                + "What can I do for you?";
+    }
+
+    /**
+     * Prints out goodbye message.
+     */
+    public String showGoodbye() {
```
TZL0: Is it better to name it getGoodbyeMessage?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public void printOut(String str) {
+        String line = "____________________________________________________________\n";
+        System.out.println(line + str + "\n" + line);
+    }
+
+    /**
+     * Prints out welcome message.
+     */
+    public String showWelcome() {
```
TZL0: Is it better to name it getWelcomeMessage?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,126 @@
...
+        return "I don't know this command. Try another one!";
+    }
+
+    /**
+     * Processes commands related to creating tasks.
+     *
+     * @param input The user's input.
+     * @param type The type of task.
+     */
+    public String processTask(String input, String type) {
```
TZL0: Is it better to name it parseTask()?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,117 @@
...
+        this.storage.update(tasks);
+    }
+
+    /**
+     * Returns a list of tasks which contain the given input in their descriptions.
+     *
+     * @param input The String to search for.
+     * @return String containing a numbered list of the results.
+     */
+    public String find(String input) {
```
TZL0, line 78: Does it really returns a List of Tasks?

src/main/java/duke/Ui.java
cfyjoanna: Nice suggestion! Will make this and the other changes.

src/main/java/duke/TaskList.java
cfyjoanna: The intended meaning was a text containing all the results found in String format (in numbered list form), and not the List object. Is there another way I should phrase this to make it clearer? Thanks!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+                }
+                tasks.add(task);
+            }
+            reader.close();
+        } catch (FileNotFoundException e) {
+            // Create the empty file if there is no existing file.
+            File file = new File(filePath);
+            try {
+                file.getParentFile().mkdirs();
+                file.createNewFile();
```
RichDom2185: Do you think it would be better if we declare `file` in the same scope as the `try` block?

```suggestion
            try {
                File file = new File(filePath);
                file.getParentFile().mkdirs();
                file.createNewFile();
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
...
+     * @param tasks New tasks to be updated.
+     */
+    public void update(ArrayList<Task> tasks) {
+        try {
+            BufferedWriter writer = new BufferedWriter(new FileWriter(filePath));
+            for (int i = 0; i < tasks.size(); i++) {
+                String str = tasks.get(i).toString();
+                StringBuilder task = new StringBuilder();
+                task.append(str.charAt(1) + " // ");
+                if (str.charAt(4) == 'X') {
```
RichDom2185: Maybe constants could be used here instead of so-called "magic numbers" for clarity?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Returns true if task description contains the given input.
+     *
+     * @param input The String to search for.
+     * @return Boolean if the input is found.
+     */
+    public boolean contains(String input) {
+        return this.description.toLowerCase().contains(input.toLowerCase());
```
RichDom2185: Would it be better if we avoid the unnecessary use of the `this` keyword, since `description` is not shadowed?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Marks task as done.
+     */
+    public void markAsDone() {
+        this.isDone = true;
```
RichDom2185: Ditto on `this` and shadowing.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Marks task as undone.
+     */
+    public void markAsUndone() {
+        this.isDone = false;
```
RichDom2185: Ditto on `this` and shadowing.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Returns details of the task.
+     *
+     * @return Details of task.
+     */
+    @Override
+    public String toString() {
+        return "[" + this.getStatusIcon() + "] " + this.description;
```
RichDom2185: Ditto on `this` and shadowing. The issue is also present in a couple other places in the code.

RichDom2185 (commented):
Overall, looks really good. Great attention to the coding standard!

src/main/java/duke/Storage.java
cfyjoanna: Thanks for pointing these out! Will rectify.

PR #94, opened 2022-08-20
[Richard Dominick] iP

RichDom2185:
# Mia

Mia is an assistant that has some of the following quality-of-life updates to the CLI:
* All messages are wrapped in chat bubbles 💬 
* Automatically breaks long inputs across multiple lines
* _...and many more..._

How to run:
1. Make sure you have Java 11 installed
1. Head over to the [`master`](/RichDom2185/ip/tree/master) branch and find download the latest `.jar` file from the "Releases" tab
1. Run the `.jar` file

~~Make sure you run it from a normal terminal window, not IntelliJ's integrated terminal.~~ In order to support ANSI escape codes (to overwrite the text prompt with a chat bubble), please **use a terminal emulator that supports it**.

Todo:
- [x] Replace `String` with `Span`
- [ ] Support more rich text types over the CLI
- [ ] Add GUI
- [ ] Deprecate CLI 😢 

## The `Text` Interface

> View the source code [here](/RichDom2185/ip/tree/master/src/main/java/general/ui/Text.java).

Originally meant to replace `String` in order to support overriding of `String::length` when using ANSI escape codes, so far only `Span` (direct substitute for a normal `String`) uses it due to lack of time to implement other classes.

```java
// ...
public interface Text {
    String getText();

    int length();

    Stream<? extends Text> lines();

    Text[] split(String regex);

    Text strip();
}
```


src/main/java/general/ui/Text.java
```diff
@@ -0,0 +1,15 @@
+package general.ui;
+
+import java.util.stream.Stream;
+
+public interface Text {
```
TZL0, line 5: I like your class Text to process String in the UI.

src/main/java/general/ui/ChatWindow.java
```diff
@@ -0,0 +1,80 @@
...
+public class ChatWindow {
+    private final int windowWidth;
+    private final Scanner sc;
+
+    public ChatWindow(int windowWidth) {
+        this.windowWidth = windowWidth;
+        sc = new Scanner(System.in);
+    }
+
+    private static Text[] breakLines(Text data) {
```
TZL0, line 22: Is it better to name it getBrokenLines()?

src/main/java/mia/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class Deadline extends Task {
+    private LocalDate byDate;
+    private LocalTime byTime;
+
+    Deadline(String title, String deadline) throws IllegalArgumentException {
+        this(title, false, deadline);
+    }
+
+    Deadline(String title, boolean isCompleted, String deadline) throws IllegalArgumentException {
```
TZL0: Is it better to add an access modifier?

src/main/java/mia/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * A {@code Deadline} is a {@code Task} that has a due date and time.
+ *
+ * @author Richard Dominick
+ */
+public class Deadline extends Task {
+    private LocalDate byDate;
+    private LocalTime byTime;
+
+    Deadline(String title, String deadline) throws IllegalArgumentException {
```
TZL0: Is it better to add an access modifier?

src/main/java/mia/Event.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @author Richard Dominick
+ */
+public class Event extends Task {
+    private LocalDate startDate;
+    private LocalTime startTime;
+    private LocalDate endDate;
+    private LocalTime endTime;
+
+    Event(String title, boolean isCompleted, String time) {
```
TZL0: Is it better to add an access modifier?

src/main/java/mia/Event.java
```diff
@@ -0,0 +1,69 @@
...
+        this(title, isCompleted, time, time);
+    }
+
+    Event(String title, String time) {
```
TZL0: Is it better to add an access modifier?

src/main/java/mia/Mia.java
```diff
@@ -0,0 +1,66 @@
...
+    public void respond(String message) {
+        window.printResponse(new Span(message));
+    }
+
+    /**
+     * Gets the {@code TaskManager} task list manager for this {@code Mia} instance.
+     *
+     * @return The {@code TaskManager} associated with this {@code Mia} instance.
+     */
+    public TaskManager getTasks() {
```
TZL0, line 40: Why not name it getTaskManager() to better identify return type?

src/main/java/general/ui/ChatWindow.java
jwdavis0200: I agree, according to the CS2103T coding standards such a getter method should have the verb infront of it :-)
Interesting work though with this method!

src/main/java/general/ui/ChatWindow.java
```diff
@@ -0,0 +1,80 @@
...
+        // Break strings at the next word after line length hits 30 characters
+        return data.lines()
+                .flatMap((s) -> Arrays.stream(s.split("(?<=\\G\\b?.{30,}\\s)")))
+                .map(Text::strip)
+                .toArray(Text[]::new);
+    }
+
+    public void dispose() {
```
jwdavis0200, line 30: Hmm, since it's just a single line method call, would it be better just to skip the implementation of this as a method since you would still have to call the method anyways. Could it affect the number of stack calls the program have to call? Although performance difference would be really negligible but something to think about.

src/main/java/general/ui/ChatWindow.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public void printResponse(Text response) {
+        final Text[] lines = breakLines(response);
+        int maxLength = 3; // prevents negative count
+        for (int i = 0; i < lines.length; i++) {
+            if (maxLength < lines[i].length()) {
+                maxLength = lines[i].length();
+            }
+        }
+        final int paddingLength = windowWidth - maxLength - 4;
```
jwdavis0200, line 47: Great use of constant here, since padding length won't be changing.

src/main/java/general/utils/StoredFile.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * A utility class that wraps abstracts the boilerplate of reading and writing to a file.
+ *
+ * @author Richard Dominick
+ */
+abstract public class StoredFile {
+    public static StoredFile from(String filePath) {
+        final File file = new File(filePath);
+        if (file.exists()) {
+            return new ExistentFile(filePath);
```
jwdavis0200, line 19: Good naming scheme, since file exists, the class is an ExistentFile!

src/main/java/mia/Command.java
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * Parses a user input string into a command that is appropriate for the calling context.
+     *
+     * @param context The context the command is intended to be executed from
+     * @param input The user input string to parse
+     * @return A {@code Command} for the context to call
+     * @throws IllegalArgumentException When a parsing error occurs
+     */
+    public static Command from(Mia context, String input) throws IllegalArgumentException {
```
jwdavis0200, line 29: No problem here, but seeing you're really proficient, why not simplify your code by implementing enums? :-)

src/main/java/mia/TaskManager.java
```diff
@@ -0,0 +1,168 @@
...
+                                case 'D':
+                                    return Stream.of(Deadline.fromSaveFormat(line.substring(3)));
+                                default:
+                                    System.out.printf("Invalid save file format: %s\nSkipping this line... please check for corrupted data.\n", line);
+                                    return Stream.empty();
+                                }
+                            }
+                            return Stream.empty();
+                        }
+                ).collect(Collectors.toList());
```
jwdavis0200, line 50: Impressive work with the Java Collections Framework!

src/main/java/mia/TaskManager.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    /**
+     * Searches all tasks for titles matching the given query.
+     *
+     * @param query The query to be word-matched against the task titles
+     * @return A {@code String} representation of the tasks that matched the given query
+     */
+    public String search(String query) {
```
jwdavis0200, line 133: Since your method will be searching AND returning the tasks, maybe searchResult would be a better name?

src/test/java/TodoTest.java
```diff
@@ -0,0 +1,20 @@
...
+        final Todo todo = Todo.fromSaveFormat(DUMMY_TODO);
+        assertEquals("T;;" + DUMMY_TODO, todo.toSaveFormat());
+    }
+
+    @Test
+    public void stringTest() {
+        final Todo todo = Todo.fromSaveFormat(DUMMY_TODO);
+        assertEquals("📜 ☐ Title", todo.toString());
+    }
+}
```
jwdavis0200, line 21: I love your extensive testing of your program, this is definitely a very good habit for future development projects!

jwdavis0200 (commented):
Really impressive work!! I have left some comments on some potential coding standard issues, really minor ones. Your program has been one that's very interesting to review. There are also comments left by another reviewer, whose comments I agree with. Otherwise LGTM, great job!

src/main/java/general/ui/Text.java
RichDom2185: Thanks! I created the interface mainly because I had envisioned for proper rich text support, so whatever that is passed in may not be a pure `String`, so I wanted to make a common standard `Text` that was understandable by the app as something to parse and print accordingly

src/main/java/general/ui/ChatWindow.java
RichDom2185: Hmm, could you point me to which part specifically? This isn't a getter method by any means, as it takes in a `Text` and returns a broken-up version of it. It does not depend on any internal object state at all.

src/main/java/mia/Deadline.java
RichDom2185: Good catch, will add!

src/main/java/mia/Deadline.java
RichDom2185: Good catch, will add!

src/main/java/mia/Event.java
RichDom2185: Good catch, will add!

src/main/java/mia/Mia.java
RichDom2185: Because I felt that the return type should not influence the method name. In my opinion, the method name should return what the object _is_ (e.g. "get my laptop"), in natural language (to be as clear as possible), as opposed to specifying the exact specification (e.g. "get my HP Book Series-12403", as this is provided from the method signature anyway.

What do you think?

_PS: The HP model is fake (I think)_

src/main/java/general/ui/ChatWindow.java
RichDom2185: This (using a wrapper method) is the only way to call it without exposing the entire internals of how `ChatWindow` works. In terms of number of stack calls, I would argue that leaving the resource (in this case, `Scanner`) open when we're no longer using it is worse than an extra function call. Plus, compiler does its own optimisations. What do you think?

src/main/java/general/ui/ChatWindow.java
RichDom2185: Thanks!

src/main/java/general/utils/StoredFile.java
RichDom2185: Thanks!

src/main/java/mia/Command.java
RichDom2185: I don't see the need for having a separate enum for every single command type. I like to think of it as: what if in the future you want to have another command that exits while not saving any changes (for example)? What about other behaviour? It would not make sense to use multiple conditional checks for every single enum type.

But by using "flags" such as `shouldCloseContext`, or potentially `shouldXXXX` in the future, this makes it easier. Also using enums breaks abstraction. Why should a the `context` the command is being run in know about what types of `Command`s are available?

src/main/java/mia/TaskManager.java
RichDom2185: Thanks!

src/main/java/mia/TaskManager.java
RichDom2185: Hmm I'm not too sure what you are trying to convey, as I don't really find `searchResult` that natural-sounding. The method name should start with a verb right, so `searchSomething`. Maybe you mean `searchTasks`?

src/test/java/TodoTest.java
RichDom2185: Thanks!

PR #108, opened 2022-08-21
[Lee Ian Ee] iP

leeianee:
# iP Pull Request

- Bullet1
- Bullet2

1. Numbered 1
2. Numbered 2

```java
String s = "Fenced code block";
```

- [x] Task List Item 1
- [ ] Task List Item 2

😄 

> Block quote

This is a in-line [Hyperlink](https://github.com/)

**Bold**


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Constructor for Deadline.
+     * @param description Description of the task.
+     * @param by When the task has to be done by.
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+        try {
+            LocalDate d1 = LocalDate.parse("2019-12-01");
```
kevinchangjk: Hmm I'm not really sure but this doesn't seem to belong here? Maybe it was left here by mistake?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Starts the Duke bot.
+     */
+    public void run() {
+        Ui.startMessage();
+        Parser parser = new Parser(this.tasks);
+        String input = "";
+        Scanner scan = new Scanner(System.in);
+        input = scan.nextLine();
```
kevinchangjk: I think you could declare and instantiate `input` on the same line here, to make it more concise.

```suggestion
        String input = scan.nextLine();
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,52 @@
...
+
+        while (parser.parse(input)) {
+            input = scan.nextLine();
+            storage.writeToFile(tasks);
+        }
+        Ui.endMessage();
+    }
+
+    public static void main(String[] args) {
+        new Duke("data/duke.txt").run();
```
kevinchangjk: In this case, the filepath `"data/duke.txt"` is hard-coded right? Perhaps it can be declared earlier on as a constant?

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Constructor for DukeException for an invalid input.
+     */
+    public DukeException() {
+        super();
+        this.description = "☹ OOPS!!! I'm sorry, but I don't know what that means :-(";
+    }
+
+    /**
+     * Constructor for DukeException for a missing task descriptor.
```
kevinchangjk, line 18: Minor nit, but I think there should be an empty line between the descriptor and params sections here? I noticed that you added them for some other comments, so maybe can be more consistent with the javadocs comments.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Constructor for Parser.
+     *
+     * @param list
+     */
+    public Parser(TaskList list) {
+        this.list = list;
+    }
+
+    /**
```
kevinchangjk, line 34: I think this javadocs comment is missing a method descriptor?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+                        break;
+                    case "find":
+                        if (split.length < 2) {
+                            throw new DukeException("find");
+                        } else {
+                            String keyword = split[1];
+                            this.list.findTask(keyword).printTaskList();
+                        }
+                        break;
+                    default: // Default case
```
kevinchangjk: This is based on what I understand so far from your code.

You have a set of valid inputs declared earlier, and here you use the switch case statement to check which valid command the input matches, but this is only after an if statement basically checking if the first word matches any valid commands.

If so, then logically (assuming your program works as it should) the default case should never occur right? 

If that is so, I think it would make more sense to simply have it be empty, and instead, at **line 91** below (reached after there is an invalid input), the error message can be included as `"Wrong input"`, instead of being blank. Would that be better?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,99 @@
...
+                        System.out.println("Wrong input");
+                    }
+                } else {
+                    throw new DukeException(); // Invalid input
```
kevinchangjk: As mentioned above, maybe something like this?

```suggestion
                    throw new DukeException("Wrong input"); // Invalid input
```

kevinchangjk (commented):
Quite detailed and thorough work. I think this is great as a product, but perhaps the code standards can be a bit more consistent. Good job!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Constructor for Parser.
+     *
+     * @param list
+     */
+    public Parser(TaskList list) {
+        this.list = list;
+    }
+
+    /**
```
Chustinjeng, line 34: missing method descriptor in javadocs

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Adds an Event object in the TaskList.
+     * @param description The description of the Event object.
+     * @param when When the Event will occur.
+     */
+    public Task addEvent(String description, String when) {
+        this.list.add(new Event(description, when));
+        return this.list.get(list.size() - 1);
+    }
+
```
Chustinjeng, line 97: Can perhaps merge into one addTask method? Both methods have the same parameter types so can maybe use polymorphism

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Constructor for Deadline.
+     * @param description Description of the task.
+     * @param by When the task has to be done by.
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+        try {
+            LocalDate d1 = LocalDate.parse("2019-12-01");
```
Chustinjeng, line 23: May be a bug here, but nothing serious :)

Chustinjeng (commented):
Nice work! Overall I felt your code was well-organised and easy to read, but I guess there are some careless mistakes but it happens 👍

PR #112, opened 2022-08-21
[Neo Sun Han] iP

neosunhan:
# Duke

> "Best app ever!" – a satisfied user

- text-based
- easy to learn

All you need to do is,
1. download duke.jar from [here](https://github.com/neosunhan/ip/releases/tag/v0.1)
2. copy the jar file into an empty folder
3. open a command window in that folder
4. run the command `java -jar duke.jar`

And it is **FREE**! :+1:

Features:
- [X] Managing tasks
- [ ] Reminders (coming soon)

Prerequisites: JDK 11

## Using the `Duke` class

The `Duke` constructor takes in a `java.nio.file.Path` object that represents
the path to the storage file (the text file used to store tasks). Users can then
call the `run` method to start Duke.

```java
Path storagePath = Paths.get(System.getProperty("user.dir"), "data", "duke.txt");
new Duke(storagePath).run();
```


src/main/java/duke/command/AddEventCommand.java
```diff
@@ -0,0 +1,36 @@
...
+import duke.task.Task;
+import duke.task.TaskList;
+
+/**
+ * Command to add an Event to the list of tasks
+ */
+public class AddEventCommand extends Command {
+    public static final String COMMAND_WORD = "event";
+
+    private final String desc;
```
xav168, line 16: hmm shouldn't desc be DESC since its a constant?

src/main/java/duke/command/AddEventCommand.java
```diff
@@ -0,0 +1,36 @@
...
+    private final LocalDate at;
```
xav168, line 17: similarly, shouldn't at be AT?

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,36 @@
...
+
+import duke.storage.Storage;
+import duke.task.Deadline;
+import duke.task.Task;
+import duke.task.TaskList;
+
+/**
+ * Command to add a Deadline to the list of tasks.
+ */
+public class AddDeadlineCommand extends Command {
```
xav168, line 13: hmm perhaps could consider a different name for the class.

xav168 (commented):
LGTM, just a few nits here and there

PR #128, opened 2022-08-21
[Chen Guanzhou] iP

Guanzhou03:
- ~~FAST~~ SUPER FAST to use
- Easy to learn

Use these easy steps:
1.  download from [here](https://github.com/Guanzhou03/ip/releases)
2.  double click
3.  add your tasks
4.  enjoy using the task manager!

AND IT IS **_FREE!_** :star_struck:

Enjoy these features: 
- [ ] Managing tasks
- [ ] Reminders (coming soon)

Here's the `main` method:

```java
public class Duke {
        Actions.toDoList();
    }
}
```


src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,14 @@
+package duke;
+
+/**
+ *  A class which encapsulates the exceptions that Duke throws
+ *  @author  Chen Guanzhou
+ *  @version v1
+ */
+
```
hanwenlai: Perhaps no blank line between the documentation block and the class?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,44 @@
...
+ *  A class which encapsulates the user interactions between Duke and the user.
+ *  @author  Chen Guanzhou
+ *  @version v1
+ */
+public class Ui {
+
+    /**
+     * A method for Duke to tell the user how many tasks they currently have.
+     */
+    public static String taskNumberMessage(TaskList taskList) {
```
hanwenlai: Perhaps the method name can be getTaskNumberMessage()?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Adds a task to the end of the task list.
+     * @param t The task to be added.
+     */
+    public void addTask(Task t) {
+        this.taskList.add(t);
+    }
+
+    
+    public TaskList matchingItems(String input) {
```
hanwenlai: Should the method name be getMatchingItems()? Besides, perhaps add a header comment to this method?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,44 @@
...
+        else {
+            return "Now you have " + taskList.getLength() + " tasks in the list.";
+        }
+    }
+
+    /**
+     * Greeting message when Duke is first executed.
+     * @return
+     */
+    public static String greetingMessage() {
```
hanwenlai, line 27: Perhaps this method name can be getGreetingMessage()?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,44 @@
...
+     * Error message when file is not found when trying to read from a file.
+     */
+    public static void handleFileNotFoundException() {
+        System.out.println("File not found!");
+    }
+
+    /**
+     * Ending message after user terminates the program with "bye"
+     */
+    public static String endingMessage() {
```
hanwenlai, line 42: Should this method name be getEndingMessage()?

hanwenlai (commented):
Well done. 👍 Except for the naming of some methods, code generally looks good. Coding standards are also mostly obeyed.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,130 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+
+/**
+ *  A class that handles parsing and processing of user inputs.
+ *  @author  Chen Guanzhou
+ *  @version v1
+ */
+public class Parser {
+    private TaskList currList;
```
ekweirui, line 24: Perhaps it would be more appropriate if tasklist was not an attribute of parser as a parser does not have a tasklist ?🤔

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,130 @@
...
+                result += "Here are the matching tasks:\n";
+                TaskList matched = currList.matchingItems(searchWord);
+                for (int i = 0; i < matched.getLength(); i++) {
+                    result += i + 1 + ". " + matched.getTaskAt(i).toString() + "\n";
+                }
+                return result;
+            default:
+                return result;
+            }
+
```
ekweirui: The block of code feels heavy, perhaps the logic could be spread out elsewhere, with only the key idea such as the case-switch statement being here 🤔 .

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
+package duke;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Scanner;
+import java.nio.file.Path;
```
ekweirui, line 15: It would be neater if the import statements were sorted alphabetically?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,110 @@
+package duke;
```
ekweirui, line 1: Adding a line break to separate the package statement from the import statements might add clarity! Similar for other classes.

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,14 @@
+package duke;
```
ekweirui: Perhaps it would be better to group all the exceptions in a package?

ekweirui (commented):
Overall, your code looks really good and clean! I learnt quite a bit from reading your code, and definitely think you are on the right track! Other than a few small suggestions, I think you did a great job! 😃

src/main/java/duke/DukeException.java
Guanzhou03: Thank you

src/main/java/duke/Ui.java
Guanzhou03: Good point!

src/main/java/duke/TaskList.java
Guanzhou03: Good points!

src/main/java/duke/Ui.java
Guanzhou03: I agree

src/main/java/duke/Parser.java
Guanzhou03: Good point with the has-a relationship, will think about changing this

src/main/java/duke/Parser.java
Guanzhou03: Definitely possible to abstract out the logic, thanks for the suggestion.

src/main/java/duke/Storage.java
Guanzhou03: Thank you

src/main/java/duke/Storage.java
Guanzhou03: Thank you

shamooose (commented):
Good organisation of your code!

PR #158, opened 2022-08-21
[Tan Wen Cong] iP

TanWenCong:
# Duke
> Subtracting from your list of priorities is as important as adding to it. - Frank Sonnenberg, [Soul Food: Change Your Thinking, Change Your Life](https://www.goodreads.com/work/quotes/66520153)

Leave all the memory work to Duke! Duke is
- text-based
- easy to learn

All you need to do is 
1. Download it from [here](https://github.com/TanWenCong/ip)
2. Git clone
3. Run the jar file
4. One example command you can execute is `todo read` when running Duke

And it is **FREE**! :grin:

Features:
- [X] Managing Todo
- [X] Managing Deadline
- [X] Managing Event

Here is the main method, you may edit the file path to store your tasks at:

```java
    }
```


src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,27 @@
...
+     * Adds task to taskList and save changes
+     *
+     * @param taskList TaskList object containing tasks input by user before
+     * @param ui       Ui object that interacts with user
+     * @param storage  Storage object that saves to and loads from storage file
+     */
+    @Override
+    public void execute(TaskList taskList, Ui ui, Storage storage) {
+        taskList.addTask(task);
+        storage.save(taskList);
```
hauchongtang: Good to see that you have abstracted parsing and task storage logic to `TaskList` and `Storage` classes.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,19 @@
+package duke;
+
+public abstract class Command {
```
hauchongtang, line 7: Good use of abstract classes, since `Command` need not be instantiated anywhere in the code.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+        switch (taskName) {
+        case "todo":
+            return new Todo(taskDescription, isDone);
+        case "event":
+            return new Event(temp[0], temp[1].split(" ", 2)[1], isDone);
+        case "deadline":
+            return new Deadline(temp[0], temp[1].split(" ", 2)[1], isDone);
+        default:
+            // should not reach this part
+            return null;
```
hauchongtang, line 139: Try not to return null here, you could instead throw an Exception, reason being that for possible future extensions of Duke, it would be harder to debug, especially when faced with `NullPointerException`.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Scanner;
+
+public class Storage {
+    String filepath;
```
hauchongtang: Would you need `filepath` in other classes ? Else, you should consider maintaining private accessibility.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,74 @@
+package duke;
+
+public abstract class Task {
```
hauchongtang, line 7: Similar to `Command` class, good job in instantiating it as an abstract class.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+        System.out.println("Here are the matching tasks in your list:");
+        temp.showTasksPlain();
+    }
+
+    /**
+     * Shows the tasks in TaskList without additional comments
+     */
+    private void showTasksPlain() {
```
hauchongtang: Maybe you can consider renaming this function for better clarity based on what your code logic for this function is, which seems to be outputting a numbered list of tasks stored in .txt file locally.
```suggestion
    private void showNumberedTasks() {
```

hauchongtang (commented):
Overall, code was written well with good code quality. However, there are some minor stuff you should take note of. Also, it would be better if you could package your classes accordingly based on what role do they play in this chatbot for better readability and extensibility in future.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+        // split into ["userCommand", "description"]
+        String[] userInput = fullCommand.split(" ", 2);
+        String userCommand = userInput[0].toLowerCase();
+        Command command;
+
+        if (!SET_OF_COMMANDS.contains(userCommand)) {
+            throw new DukeException("I'm sorry, but I don't know what that means.");
+        }
+
+        switch (userCommand) {
```
nareus, line 30: I like that you have used indentation here appropriately

src/main/java/duke/UnmarkCommand.java
```diff
@@ -0,0 +1,27 @@
+package duke;
+
+public class UnmarkCommand extends Command {
```
nareus, line 7: I like that the names of classes and methods are following coding standards

nareus (commented):
Looks good. Coding standard seems to be in order

PR #162, opened 2022-08-21
[Lee Cheng Da] iP

chengda300:
# Dude
> It sounds like Duke, but my name is actually Dude. -👦 

This bot *can* manage a list of tasks by:
- Specifying deadlines
- Marking a task's completion status
- Filtering task names within the list of tasks

To **use** this bot:
1. Load previous run's data if any
2. Add tasks to list
3. Mark tasks as completed, or delete tasks

This bot has supported the lives of:
- [x] 1 person
- [ ] 10 people
- [ ] 100 people
- [ ] 1,000 people
- [ ] 10,000 people
- [ ] 100,000 people
- [ ] 1,000,000 people

This is how this bot is created using the `Duke` constructor method.
```java
public Duke() {
    this.ui = new Ui();
    this.storage = new Storage(this.filePath);
    this.tasks = new TaskList();
    this.parser = new Parser();
}
```

**This bot is [open source](https://github.com/chengda300/ip) too!**


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+                    d.getStorage().writeToFile(this.delete + this.space + parameter + "\n", isLoading);
+                } catch (IOException e) {
+                    System.err.println("This command could not be saved due to an unknown error.");
+                }
+            } catch (NumberFormatException e) {
+                System.err.println("Invalid task.");
+            }
+        } else if (input.startsWith(this.find + this.space)) {
+            String s = input.substring((this.find + this.space).length());
+            if (s.isBlank()) {
```
yiyefyy: I think you can avoid deep nesting by using switch.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,175 @@
...
+
+import java.io.IOException;
+import java.time.LocalDate;
+import java.util.Scanner;
+
+/**
+ * Represents the Duke bot.
+ */
+public class Duke {
+    private static boolean isLoading = true;
```
yiyefyy, line 22: Nice naming!

src/test/java/TaskTest.java
```diff
@@ -0,0 +1,25 @@
+import duke.Todo;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class TaskTest {
+    Todo t = new Todo("123");
+
+    @Test
+    void testGetStatusIcon() {
```
yiyefyy, line 11: you can use underscores to name the test methods: eg test_Find_Success.

yiyefyy (commented):
I think your code is quite good and mostly followed the coding standards.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+
+/**
+ * Represents a parser to parse inputs into actions to be taken.
+ */
+public class Parser {
+    private String bye = "bye";
```
nqt230: I like how you define the command strings as a class field. You can also consider making these constant, since the commands are not likely to change over the course of running the program (unless you support command remapping).

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Parses an input to decode it.
+     *
+     * @param d Duke bot to be instructed.
+     * @param input Input to be parsed.
+     * @param isLoading Boolean to indicate if the Duke bot is loading.
+     */
+    public void parse(Duke d, String input, boolean isLoading) {
+        if (input.equals(this.bye)) {
```
nqt230: Since there are multiple if-else statements to check the input against the possible commands, would a switch statement be better here for readability?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+                }
+            } else {
+                String task = s.substring(0, s.indexOf(this.deadlineBy) - this.space.length());
+                String deadline = s.substring(s.indexOf(this.deadlineBy) + this.deadlineBy.length()
+                        + this.space.length());
+                if (task.isBlank()) {
+                    System.out.println("The task is empty, what do you really mean?");
+                } else if (deadline.isBlank()) {
+                    System.out.println("The deadline is empty, do you mean it has no deadline?");
+                    System.out.println("If it is, please add it as a todo instead.");
```
nqt230: I like the user feedback you give here, I can see how these prompts would be very helpful for a first time user.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @param d Duke bot that is running.
+     * @param filePath Path of the file that contains saved information from the last run.
+     * @throws FileNotFoundException If file is not found.
+     */
+    public void loadData(Duke d, String filePath) throws FileNotFoundException {
+        File f = new File(filePath);
+        Scanner s = new Scanner(f);
+        while (s.hasNext()) {
+            d.parse(s.nextLine());
```
nqt230, line 38: It seems your approach to loading back the data is to save the entire command strings and parse them when you want to load the data in, reconstructing the tasks this way. I think this could potentially result in save file bloating if there are multiple repeated delete commands, although this is not a big issue given the scale of the project. Would it be better to convert Tasks into some kind of save string format that you can use to reconstruct the Task?

nqt230 (commented):
Looks good :+1:
Just some minor concerns regarding the approach to saving of data and parsing of input, but I think overall the code is good.

src/main/java/duke/Duke.java
chengda300: Thank you!

src/test/java/TaskTest.java
chengda300: Ok, I will adjust the names

src/main/java/duke/Parser.java
chengda300: Yes, it should have been private static final instead, I will change them.
Thank you!

src/main/java/duke/Storage.java
chengda300: Yes, it is a lazy way of storing and loading data for small scale projects.
For large scale projects it should undo the action as if the previous action did not happen at all to keep the list as short as possible.

This lazy way allows autosave in case the program crashes halfway. It is possible to add in an optimiser function during loading to debloat the text file.

src/main/java/duke/Parser.java
chengda300: Switch statements with the current code structure would be too complicated.
Instead, I will shift the bulk of the processes to private methods of the parser class and have the parser call those methods instead. Apparently, the code was not following the single level of abstraction principle which resulted in a very long parse method.

src/main/java/duke/Parser.java
chengda300: Switch statements with the current code structure would be too complicated.
Instead, I will shift the bulk of the processes to private methods of the parser class and have the parser call those methods instead. Apparently, the code was not following the single level of abstraction principle which resulted in a very long parse method.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * Adds a task with deadline.
+     *
+     * @param s Task description.
+     * @param d Deadline in String format.
+     * @return Associated message from Duke.
+     */
+    public String addDeadline(String s, String d) {
+        return this.tasks.addDeadline(s, d);
```
totsukatomofumi: I like the use of overloading methods!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,287 @@
...
+        return this.parser.parse(this, s, this.hasFinishedLoading());
+    }
+
+    /**
+     * Prints a list of tasks with the matching input string.
+     *
+     * @param s String to request for tasks with the matching string.
+     * @return List of tasks with matching string.
+     */
+    public String find(String s) {
```
totsukatomofumi: Would it be better if the method naming was a bit more descriptive?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,287 @@
...
+
+        AnchorPane mainLayout = new AnchorPane();
+        mainLayout.getChildren().addAll(scrollPane, userInput, sendButton);
+
+        scene = new Scene(mainLayout);
+
+        stage.setScene(scene);
+        stage.show();
+
+        //Step 2. Formatting the window to look as expected
```
totsukatomofumi, line 227: I like the comments that help explain the underlying code.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,51 @@
...
+
+//@@author chengda300
+//Reused from https://nus-cs2103-ay2223s1.github.io/website/schedule/week2/project.html
+// with minor modifications
+
+/**
+ * Represents an event. An event has a description and a time that it will happen.
+ */
+public class Event extends Task {
+    private String at;
```
totsukatomofumi: Will it be better to give this a slightly more descriptive name?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,164 @@
...
+    private static final String UNMARK = "unmark";
+    private static final String TODO = "todo";
+    private static final String DEADLINE = "deadline";
+    private static final String DEADLINE_BY = "/by";
+    private static final String EVENT = "event";
+    private static final String EVENT_AT = "/at";
+    private static final String DELETE = "delete";
+    private static final String SPACE = " ";
+    private static final String FIND = "find";
+    private static final String ERROR_MESSAGE = "Sorry, I cannot understand what you exactly mean.\n"
```
totsukatomofumi, line 26: Very good use constants for values that may be used very often in the code.

totsukatomofumi (commented):
LGTM! Overall, I am impressed with the logic of code you have made. There are some variable names that may be improved, but I am just going by the book. Good job!

src/main/java/duke/Duke.java
chengda300: Yes, I will change it to findTask, thank you

src/main/java/duke/Duke.java
chengda300: Thank you!

src/main/java/duke/Event.java
chengda300: Yes it will be, it should be changed to event time I guess.

src/main/java/duke/Duke.java
chengda300: Actually that was in the GUI guide.

PR #182, opened 2022-08-22
[Max Tan] iP

maxtance:
# AnyaPro

- text-based
- easy to learn

All you need to do is,
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)
#
```
public class Main {
    }
}
```


src/main/java/anya/Anya.java
```diff
@@ -0,0 +1,187 @@
...
+        try {
+            this.ui.loadingFileMessage();
+            this.tasks = new TaskList(storage.loadFile());
+        } catch (IOException e) {
+            this.ui.errorMessage(e.getMessage() + " Creating a new database");
+            this.tasks = new TaskList();
+        } finally {
+            this.ui.loadFileSuccessMessage();
+            list(tasks);
+        }
```
njxue: Great that you added a `finally` block 👍 . I didn't think of that in my own implementation!

src/main/java/anya/AnyaException.java
```diff
@@ -0,0 +1,12 @@
...
+public class AnyaException extends Exception {
+    public AnyaException(String message) {
+        super(message);
+    }
+
+    @Override
+    public String getMessage() {
+        return super.getMessage();
+    }
+}
```
njxue, line 12: Would it be better to create child classes that extend `AnyaException`? That way, when you use the `throw` or `throws` keyword in your code, the type of exception being thrown is more explicit (eg. `UnknownAnyaCommandException`).

src/main/java/anya/Storage.java
```diff
@@ -0,0 +1,80 @@
...
+
+        File dir = new File("data/");
+        if (!dir.exists()) {
+            dir.mkdir();
+        }
+
+        File savedTask = new File(this.filePath);
+        if (!savedTask.exists()) {
+            savedTask.createNewFile();
+        }
```
njxue: Nice use of Guard Clauses!

src/main/java/anya/Ui.java
```diff
@@ -0,0 +1,111 @@
...
+        } else {
+            System.out.println("Anya has found these matching tasks in your list:");
+            for (int i = 0; i < filteredTasks.getLength(); i++) {
+                int index = i + 1;
+                System.out.println(index + ". " + filteredTasks.getTaskFromIndex(index).toString());
+            }
+        }
+        breakLine();
+    }
+}
```
njxue, line 122: Would it be better to prepend some of the methods with `get` (eg. `getLoadingFileMessage()`)? Because at first glance, the job of the `loadingFileMessage()` method may look like it is _loading_ the message instead of _retrieving_ the message.

njxue (commented):
Overall very neat code 😮 , I personally learnt some stuff from reading through your code!

src/main/java/anya/Ui.java
```diff
@@ -0,0 +1,111 @@
...
+        breakLine();
+    }
+
+    public void addTaskMessage(Task task, int taskListLength) {
+        System.out.println("Anya added: " + task);
+        System.out.println("Anya sees that you have " + taskListLength + " task(s) in your list.");
+        breakLine();
+    }
+
+    public void deleteTaskMessage(Task task) {
```
izzahaj: Perhaps a clearer method name here? E.g. `showDeleteTaskMessage`. I noticed the same issue in several other methods too.

src/main/java/anya/Anya.java
```diff
@@ -0,0 +1,187 @@
...
+     * Gets a filtered TaskList where each task contains the keyword.
+     *
+     * @param tasks A collection of Tasks.
+     * @param keyword The word that the task must contain.
+     */
+    public void find(TaskList tasks, String keyword) {
+        TaskList filteredTasks = tasks.getMatchingTasks(keyword);
+        this.ui.filteredTaskMessage(filteredTasks, keyword);
+    }
+}
```
izzahaj, line 268: Would it be better to separate the methods handling the commands into a few Command classes?

src/main/java/anya/Anya.java
```diff
@@ -0,0 +1,187 @@
...
+                        String eventDetails = Parser.parseEventDetails(userInput);
+                        Task task = new Event(taskName, eventDetails);
+                        addTask(tasks, task);
+                    } catch (ArrayIndexOutOfBoundsException e) {
+                        throw new AnyaException("The description/time of an event cannot be empty.");
+                    }
+                } else {
+                    throw new AnyaException("Anya doesn't understand this command.");
+                }
+
```
izzahaj: Any reason why you chose to use if-else statements over switch statements?

izzahaj (commented):
In general, I found your code quite readable except the main Anya class which is seems too long. Apart from the naming of some of the methods, I think you did a good job adhering to the coding standards.

PR #183, opened 2022-08-22
[Zsigmond] iP

zsiggg:
Zuke helps you remember your tasks. It 
- remembers the task list even after you close it
- is really simple to use on command line!

Why would you use Zuke?
1. You love the command-line
2. You need a simple task manager without too many extra features that apps like Notion provide
3. You're ~a geek~ _cool_ 🤓

How do you start using Zuke?
- [ ] Run the .jar file. Download it [here](https://github.com/zsiggg/ip/releases/tag/A-Jar).
- [ ] Add a task with /todo, /event, /deadline.
- [ ] Mark or unmark a task when you are done with the task.
- [ ] Close the program with /bye.

Zuke is object-oriented, with classes like `Command` and `Parser` one of the many used to encapsulate its functions. Here is the `main` method, found in the `Duke` class:
```java
    new Duke("data.txt", "data").run();
}
```


src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,159 @@
...
+        try {
+            storage.write(sj.toString());
+        } catch (IOException e) {
+            this.tasks.remove(task);
+            this.addCommands.remove(command);
+            throw new IOException("There was a problem writing the change to the file. duke.task.Task not added.");
+        }
+    }
+
+    public Task get(int i) throws IndexOutOfBoundsException {
```
ongwenpin: Perhaps its better to name the method more explicitly like getTask so there is less confusion in the future where they might be similar get methods?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,83 @@
+package duke;
+
+import duke.command.*;
```
ongwenpin: Perhaps it is better to list out explicitly what classes you have imported from package?

ongwenpin (commented):
Overall, code looks good and very well organised, LGTM!

src/main/java/duke/Parser.java
zsiggg: You're right, I should have listed out all of the commands explicitly for easier readability and future maintenance.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+        try {
+            storage.write(sj.toString());
+        } catch (IOException e) {
+            this.getTask(i).mark();
+            this.addCommands.set(i, commandString);
+            throw new IOException("There was a problem writing the change to the file. duke.task.Task not unmarked.");
+        }
+    }
+
+    public TaskList search(String keyword) {
```
zupey, line 152: Remember to add javadoc comments for the methods that you add! 😄

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,17 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class ParserTest {
+    @Test
+    public void parseWrongCommandTest() {
```
zupey:
```suggestion
    public void parse_wrongCommand_exceptionThrown() {
```
Perhaps you can use this naming convention to better indicate the function you are testing!

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,31 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class TaskListTest {
+    private final TaskList taskList = new TaskList();
+    @Test
+    public void addTest() {
```
zupey:
```suggestion
    public void add_deadline_successfullyAddedToTaskList() {
```
Can consider renaming test methods to be more descriptive 😄

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+        ArrayList<Task> matchingTasks = new ArrayList<>();
+        ArrayList<String> matchingAddCommands = new ArrayList<>();
+
+        IntStream.range(0, tasks.size())
```
zupey, line 156: I like that you also used functional programming in an overall OOP project! 👍

zupey (commented):
LGTM, except for some small things that you may have missed, such as not including javadoc.

Liked that you used streams and all the best for the iP and upcoming tP.

src/main/java/duke/TaskList.java
zsiggg: Resolved in a future branch!

src/test/java/duke/ParserTest.java
zsiggg: Thanks for pointing out the test method naming convention :)

src/test/java/duke/TaskListTest.java
zsiggg: Done!

PR #197, opened 2022-08-22
[Amadeus] ip

ama-chi:


src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Duke.Duke is a basic chat-bot stores tasks into a list.
+ *
+ * @author Chi Song Yi Amadeus
+ * @version A-MoreOOP
+ * @since 17-08-2022
+ */
+public class Duke {
+
+    /**
```
kangzongxian, line 25: Good to list down the description of the attributes!

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+     * Processes and runs user commands.
+     *
+     * @param command user command
+     * @param input   input string
+     */
+    private void executeCommand(Commands command, String input) {
+        if (command == Commands.LIST) {
+            System.out.println("    ____________________________________________________________\n");
```
kangzongxian: For consistency sake, I think it might be better to abstract out this line break into a variable, so that you can just use it!

src/main/java/Duke/Task.java
```diff
@@ -0,0 +1,108 @@
...
+        this.completed = completed;
+    }
+
+    /**
+     * Sets complete to true.
+     */
+    public void mark() {
+        completed = true;
+        System.out.printf(
+                "    ____________________________________________________________\n" +
```
kangzongxian: Likewise here, I think it would be good to abstract out the line break into a variable as well!

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,174 @@
+package Duke;
```
kangzongxian, line 1: Overall I find everything looks really good, with JavaDocs well written and good testing, nice!

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+            System.out.println("    ____________________________________________________________\n");
+            for (int i = 1; i <= tasks.getSize(); i++) {
+                String index = String.format("%d.", i);
+                String entry = index + tasks.getItem(i).toString();
+                ui.showEntry(entry);
+
+            }
+            System.out.println("    ____________________________________________________________\n");
+
+        } else if (command == Commands.MARK) {
```
kangzongxian, line 177: I think that there may be a lot of different cases going on here, so it may be a good idea to leave a comment on what case each block is handling!

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,174 @@
...
+            } catch (IOException e) {
+                ui.showErrorWritingMessage();
+            }
+        } else if (command == Commands.CREATETASK) {
+            try {
+                Task newTask = tasks.createTask(input, false);
+                if (newTask != null) {
+                    storage.writeToFile(newTask);
+                }
+            } catch (DukeException.EmptyTaskException | DukeException.UnkownCommandException | DukeException.InvalidParameterException error) {
```
kynapy: Line might be too long, could try to wrap the last exception into the next line!

src/main/java/Duke/Task.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return date.
+     */
+    public String getDate() {
+        return "";
+    }
+
+    /**
+     * Creates a String representation of Duke.Task.
+     */
+    public String toString() {
```
kynapy, line 93: Good to add in @ Override when overriding methods!

kynapy (commented):
LGTM, just some nits such as long lines of code that could be wrapped to the next lines and the repeated long underscores could be stored into a variable for more concise code. Cheers 😄

PR #207, opened 2022-08-22
[Connor] iP

connlim:
# Piggy 
> “Object-oriented programming is an exceptionally bad idea which could only have originated in California.” – Edsger Dijkstra ([source](https://harmful.cat-v.org/software/OO_programming/))

## Types of Tasks
- Todo
- Deadline (supports a by date)
- Event (supports an at date)

## Features
1. Add tasks
2. Delete tasks
3. Mark tasks as done/undone
4. All tasks are loaded from and saved to disk automatically 💾
5. **FAST** to use
6. ~~Simple~~Usable interface e.g. `todo My Todo task`

## Planned Features
- [x] Find tasks
- [ ] Proper GUI
- [ ] Automatic style checking
- [ ] Varargs

## Data File Loading
```java
    new Piggy("./tasks.txt").run();
}
```


src/main/java/piggy/Piggy.java
```diff
@@ -0,0 +1,44 @@
...
+     */
+    public void run() {
+        ui.showWelcome();
+        boolean isExit = false;
+        while (!isExit) {
+            String command = ui.readCommand();
+            isExit = Parser.parse(command, ui, taskList);
+            storage.writeTasks(taskList);
+        }
+        ui.showBye();
```
mohamedsaf1: good initiative greeting "bye"

src/main/java/piggy/Storage.java
```diff
@@ -0,0 +1,126 @@
...
+    List<Task> readTasks() {
+        List<Task> tasks = new ArrayList<>();
+        Scanner sc = null;
+        try {
+            sc = new Scanner(file);
+            while (sc.hasNextLine()) {
+                String[] data = sc.nextLine().split(",");
+
+                Task task;
+                boolean done = Boolean.parseBoolean(data[1]);
```
mohamedsaf1, line 43: does not follow coding standard, better naming would be for boolean "isDone"

src/main/java/piggy/Ui.java
```diff
@@ -0,0 +1,129 @@
...
+        System.out.println("Now you have " + noRemainingTasks + " tasks in the list.");
+    }
+
+    /**
+     * Show that a task has been removed along with the remaining tasks.
+     *
+     * @param task             The removed task to show.
+     * @param noRemainingTasks The number of remaining tasks.
+     */
+    void showTaskRemoved(Task task, int noRemainingTasks) {
```
mohamedsaf1: variable "noRemainingTasks" sounds like there are no remaining tasks rather than number of remaining tasks.

peppapighs: Consider renaming this method to `showRemoveTask` or `showRemovedTask`? The ordering of the words makes the method name seems unclear to me.

src/main/java/piggy/Ui.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    /**
+     * Show that a task has been added along with the remaining tasks.
+     *
+     * @param task             The added task to show.
+     * @param noRemainingTasks The number of remaining tasks.
+     */
+    void showTaskAdded(Task task, int noRemainingTasks) {
```
peppapighs: Consider renaming this method to `showAddTask` or `showAddedTask`? The ordering of the words makes the method name seems unclear to me.

src/main/java/piggy/Parser.java
```diff
@@ -0,0 +1,74 @@
...
+                    } else if (!command.contains("/at")) {
+                        throw new DukeException("An event must contain an /at");
+                    }
+                    String[] split = command.substring(6).split(" /at ");
+                    task = new Event(split[0], split[1]);
+                } else {
+                    throw new DukeException("I'm sorry, but I don't know what that means :-(");
+                }
+                taskList.add(task);
+                ui.showTaskAdded(task, taskList.size());
```
peppapighs: Great uses of return early pattern to reduce nested statements.

peppapighs (commented):
Overall, the code looks good to me. Just some minor method naming that I feel are a bit unclear.

src/main/java/piggy/Storage.java
connlim: Agreed, will be fixed.

src/main/java/piggy/Ui.java
connlim: Agreed

connlim: Yeah that makes more sense thanks

PR #208, opened 2022-08-22
[Chng Ian] iP

chngchngchng:

- text-based
- easy to learn

All you need to do is,

2. double-click it.
3. add your tasks.

And it is __FREE__!

Features:

- [X] Managing tasks
- [ ] Reminders (coming soon)

If you are a  Java programmer, you can use it to practice Java too. Here's the `main` method:

```java
public class Main {
    }
}
```


src/main/java/duke/parser/DukeParser.java
```diff
@@ -0,0 +1,235 @@
...
+        // Split rest of input string for further parsing
+        // If there is no rest of input string, then restOfInputString will be "".
+        this.restOfInputString = newString.substring(this.keyword.length()).trim();
+
+    }
+
+    /**
+     * Executes a loaded and parsed instruction.
+     *
+     * @param st Storage that we want to save data to after the instruction has been executed
```
johnbenedictyan: Perhaps this param was changed? Now, should this param tag be needed?

src/main/java/duke/parser/DukeParser.java
```diff
@@ -0,0 +1,235 @@
...
+     * Handles numerical instructions that act on an indexed element in taskList
+     *
+     * @throws DukeException Index is invalid
+     */
+    public Command numericalInstructionHandler() throws DukeException {
+
+        // First, try to parse the numerical part of the instruction, if error, throw it
+        int instructionNum;
+        try {
+            instructionNum = Integer.valueOf(this.restOfInputString) - 1;
```
johnbenedictyan: Any reason why you used `Integer.valueOf` instead of `parseInt`?

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,145 @@
...
+    public String getWriteString() {
+        String result = "";
+        for (Task task: this.taskList) {
+            result += task.getSaveString() + "\n";
+        }
+        return result;
+    }
+
+    @Override
+    public String toString() {
```
johnbenedictyan, line 156: Perhaps you could use StringBuilder in this instance to help with performance
```
int counter = 1;
StringBuilder res = new StringBuilder("Here are the tasks that you have added to the list:\n");
for (Task task : this.taskList) {
    if (task != null) {
        res.append(counter).append(". ").append(task);
        if (counter != taskList.size()) {
            res.append("\n");
        }
        counter++;
    }
}
return counter == 1 ? "There are no tasks in your task list at the moment!" : res.toString();
```

README.md
```diff
@@ -1,4 +1,4 @@
-# Duke project template
+# duke.Duke project template
```
leecaregene, line 1: Is this change unintentional?
```suggestion
# Duke project template
```

src/main/java/duke/parser/DukeParser.java
```diff
@@ -0,0 +1,235 @@
+package duke.parser;
+
+import java.time.format.DateTimeParseException;
+
+import duke.commands.*;
```
johnbenedictyan, line 16: Perhaps the imported classes should be listed explicitly?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,56 @@
+package duke;
+
+import java.util.Scanner;
```
johnbenedictyan, line 3: Maybe this unused import should be deleted, I noticed the same issue in other places too.

src/main/java/duke/parser/DukeParser.java
```diff
@@ -0,0 +1,235 @@
...
+        this.taskList = taskList;
+    }
+
+
+    /**
+     * Handles user input, and preps parser for instruction execution.
+     *
+     * @param inputString The string that we would like to parse.
+     */
+    public void parseInput(String inputString)  {
```
leecaregene, line 57: Extra space?
```suggestion
    public void parseInput(String inputString) {
```

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,120 @@
...
+                    break;
+                case "DEADLINE":
+                    newList.add(new Deadline(description, isDone, dateTimeInfo));
+                    break;
+                case "EVENT":
+                    newList.add(new Event(description, isDone, dateTimeInfo));
+                    break;
+                default:
+                    throw new DukeException("Error reading from file: " + filePath + ": Unknown " +
+                            "task type");
```
leecaregene, line 91: Line break should be before an operator?
```suggestion
                    throw new DukeException("Error reading from file: " + filePath + ". Perhaps "
                    + "the data is stored in an incorrect manner?");
                }
                String type = parsedData[0];
                boolean isDone;
                try {
                    isDone = Boolean.parseBoolean(parsedData[1]);
                } catch (Exception e) {
                    throw new DukeException("Error reading from file: " + filePath + ": Unknown "
                    + "boolean format for stored tasks");
                }

                // TODO: More error checking for description and datetimeinfo?
                String description = parsedData[2];
                String dateTimeInfo = parsedData[3];

                switch (type) {
                case "TODO":
                    newList.add(new Todo(description, isDone));
                    break;
                case "DEADLINE":
                    newList.add(new Deadline(description, isDone, dateTimeInfo));
                    break;
                case "EVENT":
                    newList.add(new Event(description, isDone, dateTimeInfo));
                    break;
                default:
                    throw new DukeException("Error reading from file: " + filePath + ": Unknown "
                    + "task type");
```

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,145 @@
...
+     * @return String that contains all matching tasks, formatted.
+     */
+    public String search(String[] terms) {
+        String result = "";
+        for (Task task: this.taskList) {
+            for (String term: terms) {
+                if (task.toString().toLowerCase().contains(term.toLowerCase())) {
+                   if (!result.contains(task.toString())) {
+                       result += task.toString() + "\n";
+                   }
```
leecaregene, line 108: Indentations off by one space?
```suggestion
                    if (!result.contains(task.toString())) {
                        result += task.toString() + "\n";
                    }
```

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param isDone Marks whether task has been completed
+     */
+    public Task(String description, boolean isDone) {
+        this.description = description;
+        this.isDone = isDone;
+    }
+
+
+    public String getStatusIcon() {
+        return isDone? "X" : " ";
```
leecaregene, line 88: Missing space?
```suggestion
        return isDone ? "X" : " ";
```

leecaregene (commented):
Some coding standard violations, you can use checkstyle to find the rest.

src/main/java/duke/parser/DukeParser.java
chngchngchng: Yep, I should remove this param.

src/main/java/duke/parser/DukeParser.java
chngchngchng: Seemed to be the first option that came to mind! Is there any difference in efficiency in using one over the other?

src/main/java/duke/tasklist/TaskList.java
chngchngchng: Good idea! Thanks for your input.

src/main/java/duke/parser/DukeParser.java
johnbenedictyan: `Integer.valueOf` actually uses `parseInt` internally, but it returns a type of Integer rather than int. I just thought for your case, `parseInt` was good enough. No major performance issues if I recall correctly.

```
public static int parseInt(String s) throws NumberFormatException {
    return parseInt(s, 10);
}

public static Integer valueOf(String s, int radix) throws NumberFormatException {
    return Integer.valueOf(parseInt(s, radix));
}

public static Integer valueOf(String s) throws NumberFormatException {
    return Integer.valueOf(parseInt(s, 10));
}
```

PR #211, opened 2022-08-22
[Li Po Hsien] iP

lpohsien:
# Duke
This is a simple to-do list manager.
## Features
Here are the commands you can use:
- add _task_
- remove _task_number_
- mark _task_number_
- unmark _task_number_
- list 

For the `add` commands, there are 3 types of tasks you can add to the to-do list. The possible tasks type are listed here:
1. **Todo**: A simple task with only a description.
2. **Deadline**: A simple task with a description and a due date.
3. **Event**: A simple task with a description, a start time, and an end time.

The encoding for the 3 task types is done based on the 
``` java
private String taskTypeString() {
    switch (this.type) {
    case TODO:
        return "T";
    case EVENT:
        return "E";
    case DEADLINE:
        return "D";
    default:
        return "";
    }
}
``` 
To use the manager, just follow these steps:
- [ ] Download the [jar file](https://github.com/lpohsien/ip/releases/tag/A-Jar)
- [ ] Launch it using `java -jar duke.jar`
- [ ] Start adding your tasks today! :+1:

> "Each day I will accomplish one thing on my to-do list." – Anonymous


src/main/java/Duke/Command/Command.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public abstract class Command {
+
+    /** Returns true if this command is the exit command.
+     * 
+     * @return True if this is the exit command. False otherwise.
+     */
+    public boolean isExit() {
+        return false;
+    }
```
Devanshshah1309: Hey, is there a reason the default implementation returns `false`? Might be better to just leave it as an abstract method right?

src/main/java/Duke/Command/DelCommand.java
```diff
@@ -0,0 +1,40 @@
+package Duke.Command;
+
+import Duke.TaskList;
+import Duke.Ui;
+import Duke.FileStorage.Storage;
+import Duke.Task.Task;
+
+/**
+ * This class represents the delete command that removes tasks
+ * from the todo list.
```
Devanshshah1309: Can I check whether "todo list" refers to the "task list" instead?

src/main/java/Duke/Command/InvalidCommandException.java
```diff
@@ -0,0 +1,12 @@
+package Duke.Command;
+
+/** 
+ * This class represents the exceptions that occurs when an
+ * invalid command was given by the user.
+ */
+public class InvalidCommandException extends Exception{
```
Devanshshah1309: Great naming for this haha, I really love the OOP design in general :)

src/main/java/Duke/Parser.java
```diff
@@ -0,0 +1,133 @@
...
+     * @return
+     * @throws InvalidCommandException
+     */
+    public static Command parse(String input) throws InvalidCommandException {
+        String errMsg = "";
+        if (input.startsWith("bye")) {
+            return new ExitCommand();
+        } else if (input.startsWith("list")) {
+            return new ListCommand();
+        } else if (input.startsWith("mark")) {
```
Devanshshah1309: Would it be better to use a switch statement here? Not sure if that would be cleaner

src/main/java/Duke/Parser.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Checks that the string can be splitted and split it to fit the command.
+     * 
+     * @param input The user input line to be delimited.
+     * @param delimiter The delimiter used to delimit the command.
+     * @param errMsg The error message associated with the given command.
+     * @return The splitted string.
+     * @throws InvalidCommandException
+     */
+    private static String[] splitWithFormat(String input, String delimiter, String errMsg)
```
Devanshshah1309: Good abstraction here haha 👍 

Can possibly abstract away the max number of splits too (e.g. in this case, 2)

Devanshshah1309 (commented):
Awesome work in general! Great code design 😄

PR #218, opened 2022-08-22
[Reuben Chay] iP 

ReubenChay:
> "Look at me I'm Mr Meseeks!" - Mr Meseeks 
# **Mr Meseeks**

Mr Meseeks is a revolutionary task manager that will improve your life! It's: 
- Text-Based using Command Line Interface (CLI) 
- User-Friendly 
- ~~Fast~~ SUPER FAST to use! 
- ~~S$1 ONLY~~ now **FREE** to use 👍 

Managing your tasks can be easy with **Mr Meseeks**, just follow these steps: 
1. download it [here](https://github.com/ReubenChay/ip/releases/tag/A-Release)
2. move the file to a folder (or anywhere which is convenient for you) 
3. open the file 
4. start using Mr Meseeks to manage your tasks for you! 😸 

Features of Mr Meseeks: 
- [x] Manage your tasks for you 
- [x] Manage Deadlines, To-Do tasks, Events 
- [x] User Interface  

Some examples of commands: 
``` 
list 
todo CS9999 Assignment 
delete 1 
mark 1 
```

If you are a Java programmer, you can use it to practice as well! Here's the main method: 
```ruby 
        new Duke().run(); 
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,275 @@
...
+import java.util.Scanner;
+import java.io.File;
+import java.io.FileWriter;
+
+
 public class Duke {
-    public static void main(String[] args) {
+    public static void main(String[] args) throws DukeException, IllegalArgumentException {
+
+        List<Task> list = new ArrayList<>(); // to store list of inputs
```
wrewsama: I think the name of this list could be more descriptive

src/main/java/Duke.java
```diff
@@ -1,10 +1,275 @@
...
+
+        Scanner sc = new Scanner(System.in);
+        String input = "";
+
+        while (true) {
+
+            try {
+                input = sc.nextLine();
+
+                if (input.equals("bye")) {
```
wrewsama: A switch case could be used here instead of if-elses

src/main/java/Duke.java
```diff
@@ -1,10 +1,275 @@
...
+     * Initialise the list from the .txt file
+     * @param list list storing all tasks 
+     */
+    static void listInit(List<Task> list) {
+        try {
+            List<String> tasks = Files.readAllLines(Path.of("./tasks.txt"));
+            for (String x : tasks) {
+                String taskType = x.substring(1,2);
+                Boolean done = x.substring(4,5).equals("X");
+                String[] arr = x.substring(6).split("\\(");
```
wrewsama: I think this name could be more descriptive

wrewsama (commented):
Some names could be more descriptive but overall LGTM!!

src/main/java/Duke.java
ReubenChay: alright, thanks!

src/main/java/Duke.java
ReubenChay: that makes a lot more sense, dont know why i didn't think of that. Thank you!

PR #228, opened 2022-08-22
[Farisa] iP

farisanadia:
# Duke
> An all-in-one tool to keep track of all the tasks in your life.  

Duke is:
- **easy** to learn
- **text based**
- **fast** to use

All you need to do is,
1. download it from [here](https://github.com/farisanadia/ip)
2. double-click it
3. add your tasks
4. let it manage your tasks for you 🙃 
And it is _FREE_

Features:
- [x] todo support
- [x] deadlines support
- [x] events support
- [x] removing & adding tasks
- [x] remembering tasks inputted previously
- [ ]  managing deadlines (coming soon)
- [ ] reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's the `main`  method:
```ruby
public static void main(String[] args) throws IOException {
        new Duke().run();
    }
```

### DIFF
--- a/src/main/java/Deadlines.java
+++ b/src/main/java/Deadlines.java
@@ -1,0 +1,21 @@
+public class Deadlines extends Task {
+    protected String taskType = "[D]";
+    protected String deadline;
+
+    public Deadlines(String description, String d) {
+        super(description);
+        this.deadline = d;
+    }
+
+    @Override
+    public void fullDesc() {
+        System.out.println("      " + this.taskType + this.getStatusIcon() +
+                " " + this.description + " (" + this.deadline + ")");
+    }
+
+    @Override
+    public String stringDesc() {
+        return this.taskType + this.getStatusIcon() +
+                " " + this.description + " (" + this.deadline + ")";
+    }
+}
--- a/src/main/java/Deadlines.java
+++ b/src/main/java/Deadlines.java
@@ -18,4 +18,10 @@ public class Deadlines extends Task {
         return this.taskType + this.getStatusIcon() +
                 " " + this.description + " (" + this.deadline + ")";
     }
+
+    @Override
+    public String textDesc() {
+        return "D : " + this.getStatusIcon() + " : " +
+                this.description + " : " + this.deadline + "\n";
+    }
 }
--- a/src/main/java/Duke.java
+++ b/src/main/java/Duke.java
@@ -1,34 +1,54 @@
-import java.util.ArrayList;
-import java.util.List;
+
 import java.util.Scanner;
 
 public class Duke {
+
     public static void main(String[] args) {
         Scanner sc = new Scanner(System.in);
-        List<String> taskList = new ArrayList<>();
+        Task[] taskList = new Task[100];
+        int indCount = 0;
         String buffLine = "    _____________________________________";
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
+
         System.out.println("    Hello from\n" + logo);
         System.out.println("    What can I do for you?\n" + buffLine);
         String userReply = sc.nextLine();
+
         while (!userReply.equals("bye")) {
-            if (!userReply.equals("list")) {
+            if (userReply.startsWith("mark")) {
+                int pos = Character.getNumericValue(userReply.charAt(5));
+                taskList[pos - 1].markAsDone();
+                System.out.println(buffLine + "\n" + "    Nice! I've marked this task as done: ");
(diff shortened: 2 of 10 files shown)


src/main/java/Deadlines.java
```diff
@@ -0,0 +1,27 @@
...
+    protected String taskType = "[D]";
+    protected String deadline;
+
+    public Deadlines(String description, String d) {
+        super(description);
+        this.deadline = d;
+    }
+
+    @Override
+    public void fullDesc() {
```
TanWenCong, line 11: ```fullDesc()``` does not not seem to imply printing the output.
Would there be a better name for this function to provide more clarity to the reader?

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+        } catch (FileNotFoundException e) {
+            System.out.println("Error reading file: file not found");
+        }
+   }
+
+   private static boolean addTaskList(String[] strArray) {
+        if (strArray[0].equals("T")) {
+            Task newTask = new ToDos(strArray[2]);
+            taskList[IND_COUNT] = newTask;
+            IND_COUNT++;
```
TanWenCong, line 37: ```IND_COUNT``` does not seem to be a constant, it might be good to use lowerCamelCase instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+            return true;
+        }
+        return false;
+   }
+
+   //rewrites entire duke.txt file when delete, mark or unmark called
+    //shld replace with code that modifies the txt file rather than rewrites it entirely
+    //is broken :'(
```
TanWenCong, line 51: Hope you fix it soon! :")

src/main/java/Deadlines.java
EmilyOng, line 11:
```suggestion
    public void getFullDesc() {
```

It would be better to [use verbs to describe method names](https://se-education.org/guides/conventions/java/intermediate.html#naming)

src/main/java/Task.java
```diff
@@ -0,0 +1,33 @@
+public class Task {
```
TanWenCong, line 1: I like how you use inheritance to reduce code duplication! Since ```Task``` does not seem to be instantiated in your code, maybe you could make it an abstract class!

src/main/java/Task.java
```diff
@@ -0,0 +1,33 @@
...
+        System.out.println("      " + this.getStatusIcon() + " " + this.description);
+    }
+
+    public String stringDesc() {
+        return this.getStatusIcon() + " " + this.description;
+    }
+
+    public String textDesc() {
+        return this.description;
+    }
```
EmilyOng, line 32: We could try to override the `toString` method of the class to retrieve the string representation of the task.

src/main/java/Deadlines.java
```diff
@@ -0,0 +1,27 @@
+public class Deadlines extends Task {
+    protected String taskType = "[D]";
+    protected String deadline;
+
+    public Deadlines(String description, String d) {
```
EmilyOng, line 5:
```suggestion
    public Deadlines(String description, String deadline) {
```

Should this [not be a single variable name](https://se-education.org/guides/conventions/java/intermediate.html#naming)?

src/main/java/Duke.java
EmilyOng, line 51: We can remove the commented code.

src/main/java/ToDos.java
```diff
@@ -0,0 +1,24 @@
+public class ToDos extends Task {
+    protected String taskType = "[T]";
+
+    public ToDos(String description) {
+        super(description);
+    }
+
+    @Override
+    public void fullDesc() {
```
TanWenCong, line 9: It seems like the child classes are not able to take advantage of this method by calling ```super.fullDesc()```, perhaps you could tweak the method in ```Task``` slightly such that you can further abstract out the code?

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+    private static int IND_COUNT;
+    private static Task[] taskList;
+
+    private static void appendToFile(String filePath, String textToAdd) throws IOException {
+        FileWriter fw = new FileWriter(filePath, true);
+        fw.write(textToAdd);
+        fw.close();
+    }
+
+   //adds previous tasks back into taskList when restarting Duke
```
EmilyOng, line 18:
```suggestion
   // adds previous tasks back into taskList when restarting Duke
```

There appears to be a [missing whitespace](https://se-education.org/guides/conventions/java/intermediate.html#layout) here. It would also be good to write this as [a JavaDoc](https://se-education.org/guides/conventions/java/intermediate.html#comments) instead.

TanWenCong (commented):
Overall structure seems okay! I like the use of inheritance, but I believe some tweaks will make your code look much cleaner. All the best in fixing your code :^)

PR #236, opened 2022-08-22
[Koh Rui Jie] iP

rjkoh:
Duke frees your mind of having to remember things you ned to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/rjkoh/ip)
2. double-click it
3. add your tasks
4. let it manage your tasks for you 💩 

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders

-----

```ruby
public class Main {
    }
}
```


src/main/java/duke/TaskList.java
```diff
@@ -11,6 +11,7 @@
 public class TaskList {
     private ArrayList<Task> tasks;
```
LuYiting0913, line 13: I like the naming of the variables

src/main/java/duke/TaskList.java
```diff
@@ -11,6 +11,7 @@
...
     private int count;
+    private static Ui ui = new Ui();
```
LuYiting0913, line 15: I like your JavaDoc. Clear and succinct

src/main/java/duke/TaskList.java
```diff
@@ -60,15 +61,15 @@ public void add(String str, Duke.Type type) {
             try {
                 tasks.add(new Todo(str));
                 count++;
-                System.out.println(String.format("Got it. I've added this task:\n" +
+                ui.printMessage(String.format("Got it. I've added this task:\n" +
                                 "%s\n" +
                                 "Now you have %d tasks in the list.",
                         tasks.get(count - 1).toString(),
                         count));
```
LuYiting0913, line 80: Nice comments for different exceptions

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,15 @@
+package duke;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+
+public class TaskListTest {
+    @Test
+    public void add_deadline_deadlineAddedWithDate() {
```
Nephelite, line 10: Nice job following naming conventions for tests

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,36 @@
+package duke;
+
+/**
+ * Deals with interactions with the user for Duke.
+ */
+public class Ui {
```
Nephelite, line 6: I dislike how this class is structured. I suggest having all of Duke's responses listed at the top of the class as constants, and then the class methods would be methods to print those constants. 

Benefits of my suggested change:
This would make editing these greetings much easier to manage as all the strings are at the top of the class.

In addition, if in the future you need to make Ui return the strings it uses in greetings for some reason, this implementation would make such a thing far easier.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,117 @@
...
+                        String dt = date + " " + time;
+                        Task event = new Event(name, LocalDateTime.parse(dt,
+                                DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm")));
+                        if (done.equals("1")) {
+                            event.mark();
+                        }
+                        tasks.add(event);
+                    } else {
+                        throw new DukeException("Unknown Task type in text file");
+                    }
```
Nephelite: Nice control flow and storage system, so I like the ideas behind it. However, I have a few suggestions to make regarding this structure.

There is far too many nested if statements. I suggest at least abstracting away the ```if(<task>.equals("<task>")``` blocks into a separate method in the task class itself, or maybe even in the ```Parser``` class as a static method. Another layer of abstraction possible is to separate the entire if block here into another method in this class, simply to reduce the levels of nested if statements and to reduce the size of this method.

The system is slightly too optimistic regarding the user. For example, if the user typos and accidentally has // in their task description.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,71 @@
...
+        // user wants to display list
+        if (command.equals("list")) {
+            return tasks.toString();
+
+        // user wants to mark certain task as done
+        } else if (command.startsWith("mark")) {
+            String str = command.replace("mark ", "");
+            int index = Integer.valueOf(str);
+            tasks.mark(index);
+            return "Nice! I've marked this task as done:\n" + tasks.getString(index);
```
Nephelite: I like how the return value is used for feedback here. Quite clever.

src/main/java/duke/MissingDescriptionException.java
```diff
@@ -0,0 +1,23 @@
+package duke;
+
+/**
+ * Exception thrown if missing name for tasks
+ */
+public class MissingDescriptionException extends DukeException {
```
Nephelite, line 6: I like that you went to create specific new exceptions for your iP.

Nephelite (changes requested):
I like many of the ideas in this iP. There are some clever bits of code and I like the effort made to create exceptions that extend from DukeException. I also like most of the method names and I think good effort has been made to ensuring coding style conventions have been met. JavaDocs look complete, and there really is not much to complain about.

The only change I would suggest is the one in the ```load()```method of ```Storage``` as it does not utilize Abstraction and overuses nested if statements thus violating coding style contentions.

PR #237, opened 2022-08-22
[Amresh] iP

amresh2000:
####Uses of Duke
- Organize your day
- Add todo tasks 
- Live updating of list

#### How to Use
1. Download it from [here] (https://github.com/nus-cs2103-AY2223S1/ip.git) 
2. double-click it
3. add ~~your~~ *all your* tasks

#### Sample code of Duke
Here is the `main` method
```
public class Main {
    }
}
```

#### Features Completed
- [x] Chatbot with basic todo, event, deadline
- [x] Reminder System
- [x] Analytics


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+import java.text.ParseException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Dealine Class to represent a class that will store Deadline object that extends from Tasks Class
+ * @author amresh A0235398R
+ */
+public class Deadline extends Task {
```
Wu-Lezheng, line 10: Maybe it be better to put classes such as Deadline.java, Event.java, etc to a separate folder called "Command" to add more clarity.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+            userInput = scanner.nextLine();
+
+            if (userInput.equals("list")) {
+                tasks.list();
+                continue;
+            }
+            if (userInput.equals("bye")) {
+                break;
+            }
+            if (userInput.split(" ", 2)[0].equals("mark")) {
```
Wu-Lezheng, line 43: What about extracting the common part `userInput.split(" ", 2)[0]` in the following `if` statement into a new variable called `commandWord`?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.List;
+
+/**
+ * Class that creates Storage object to do file manipulation
+ * @author amresh A0235398R
+ */
+public class Storage {
+
+    protected String filePath;
+    protected ArrayList<Task> storeLists = new ArrayList<>();
```
Wu-Lezheng, line 20: Maybe renaming `storeLists` which sounds like a verb to `storageList` or `tasks` which are nouns would be clearer to others?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+     * @param storeList ArrayList<Task> passed into constructor
+     * @param storage Storage object passed into constructor
+     */
+    public TaskList(ArrayList<Task> storeList, Storage storage) {
+        this.storeLists = storeList;
+        this.storage = storage;
+    }
+
+    /**
+     * Method to add task to list
```
Wu-Lezheng, line 36: Would it be better to leave a line between `description` and `@param` in JavaDoc? I notice this issue in other methods as well.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+        storage.appendToFile(todo.toString());
+
+    }
+
+    /**
+     * To allow user to create a Deadline Task
+     * @param userInput Input given by user
+     * @param by Due date for the deadline
+     */
+    public void deadline(String userInput, String by) {
```
Wu-Lezheng: Would it be better if the method name is a verb like `createDeadline` instead of a noun like `deadline`? The same applies to other methods in this class.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+        System.out.println("Now you have " + storeLists.size() + " tasks in the list.");
+        storage.appendToFile(event.toString());
+    }
+
+    /**
+     * Method to check for errors for the input that is given
+     * @param userInput Input given by the user
+     * @throws DukeException Error thrown that is specific to Duke
+     */
+    public void userInputCheck(String userInput) throws DukeException {
```
Wu-Lezheng, line 196: Would it be better to rename the method as `checkUserInput`?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,178 @@
...
+        if((firstCommand.equals("deadline") || firstCommand.equals("event")) && userInput.split("/", 2).length == 1) {
+            throw new DukeException("\t ☹ OOPS!!! The time due or needed cannot be empty.");
+        }
+    }
+
+    /**
+     * Method that activates convertToDate method
+     * @param d Deadline object passed in as parameter
+     */
+    public static void dateProcessor(Deadline d) {
```
Wu-Lezheng, line 230: Would it be better to rename the method to `activateConversionToDate` to add more clarity in terms of what the method does?

Wu-Lezheng (commented):
Overall the code looks clean. Just a few nits to fix. Maybe the whole program could be structured in another way to have greater clarity.

src/main/java/duke/Deadline.java
jacobkwan, line 10: +1 to [Lezheng's comment](https://github.com/nus-cs2103-AY2223S1/ip/pull/237/files#r959589248), but perhaps you could name that folder "task" instead of "Command". The latter would be an appropriate folder to create if you had classes like `AddCommand`, `DeleteCommand` and other `Command` abstractions which you did not implement (and that's fine).

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+    protected LocalDate date;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    /**
+     * Method to convert deadline String to LocalDate format
+     * @return boolean true value if converted
```
jacobkwan, line 27: Should there be an empty line between the description and return sections for readability? I noticed this in several of your javadocs, here's the [reference material](https://se-education.org/guides/conventions/java/intermediate.html).

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public boolean convertToDate() {
+        try {
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+            LocalDate localDate = LocalDate.parse(this.by, formatter);
+            this.date = localDate;
+            this.by = localDate.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
```
jacobkwan, line 34: Perhaps you could avoid using `this` here since the variable names are not being shadowed?
```suggestion
            date = localDate;
            by = localDate.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,178 @@
+package duke;
+
+import java.io.*;
```
jacobkwan: Perhaps it would be better to specify exactly which imports you are using instead of using the wildcard?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (IOException ex) {
+            ex.printStackTrace();
+        }
+    }
+
+    /**
+     * Method that parses strings and converts them to Tasks
+     * @param taskStr String that will be passed into method to convert
+     */
+    public void renderStringsAsTasks(String taskStr) {
```
jacobkwan: A bit of a nitpick but perhaps this method could be renamed? `render` usually has other meanings (I'm thinking 3D render etc). I like your description in the accompanying javadocs, so perhaps `convertStringToTask` would be a more intuitive name? I've removed the plurality as well (`Strings` to `String`) as the method operates on one String and returns one Task. 
```suggestion
    public void convertStringToTask(String taskStr) {
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+                fw.write("\n" + textToAppend);
+            }
+            fw.close();
+        } catch (IOException ex) {
+            ex.printStackTrace();
+        }
+    }
+
+    /**
+     * To edit Text in file primarily to mark or unmark tasks
```
jacobkwan: Should `Text` be lower case? The implication of the capital `T` is not obvious here.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,178 @@
...
+     * @param newText New text to replace old text
+     * @param oldText Old text that is meant to be replaced
```
jacobkwan: Should theses sentences be terminated with a full stop `.`? 
These minor nits occur in several of the javadocs

src/test/java/DeadlineTest.java
```diff
@@ -0,0 +1,14 @@
+import duke.Deadline;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+
+    @Test
+    public void convertToDate_success() {
```
jacobkwan, line 9: I like the naming of the test, its immediately clear what behaviour this test case is validating.

jacobkwan (commented):
LGTM, other than a few minor nits to fix! 👍 
Side note, PR description has some syntax errors
1. Header is included in the top quote (not sure if that's intended)
2. A space is required between the `#` and the header description
3. There should not be a space between your link description [click here]*no space here*(www.google.com) and the opening parenthesis of the actual link.

PR #248, opened 2022-08-22
[T Neethesh] iP

Neethesh26:
### DukePro

> "Your mind is for having ideas, not holding them." - David Allen [(source)](https://dansilvestre.com/productivity-quotes)

- text-based
- easy to learn
- ~FAST~ _SUPER_ FAST to use

All you need to do is, 

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features: 

- [X] Managing tasks
- [ ] Managing deadlines (coming soon!)
- [ ] Reminders (coming soon)

-------------------------------------------------------------------------------------------

If you are a Java Programmer, you can use it to practice Java too. Here's the `main` method: 

```
public class Main {
    }
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,82 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
```
bryanljx, line 9: Hmm, is there any reason to not provide header comments for public classes? Also, it's easy to miss out spaces like this case `Task{`, you might want to set up an actual code formatter or say intellij to auto format on save to avoid these 😄

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Information of the date of deadline.
+     */
+    private LocalDate date;
```
bryanljx, line 14: Actually there isn't a really need to write javadocs for private fields but good that you have 👍

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,52 @@
...
+                ui.showLine();
+            }
+        }
+    }
+
+
+    public static void main(String[] args) {
+        new Duke("./data/dukeInfo.txt").run();
+
+    }
```
bryanljx: Is there a reason for the empty line on line 50? It looks kind of weird to me to randomly have an empty line before the closing brace 😅

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,111 @@
...
+
+        else if (arr[0].equals("mark")){
+            if (!inputChecker(arr)) {
+                throw DukeException.MarkIndexEmptyException();
+            } else {
+                return new MarkCommand(arr);
+            }
+
+        }
+        else if (arr[0].equals("unmark")) {
```
bryanljx: I kind of understand why the if else blocks were formatted this way because of how many else blocks there are though this is still a violation of the coding standard as shown below. Perhaps a switch block would be better here?

```java
if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,131 @@
...
+                        toAdd.setTime(dateArrEvent[1]);
+                    }
+                    if (Integer.parseInt(info[1]) == 1) {
+                        toAdd.setComplete();
+                    }
+                    tasks.add(toAdd);
+                    break;
+                default:
+
+                }
```
bryanljx: An empty default case feels a bit redundant here. Would adding a comment to explain why there are no other cases/this default case be reached or simply to remove this be better?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,56 @@
...
+
+public abstract class Task {
+    /**
+     * Description of the task.
+     */
+    private String taskName;
+    /**
+     * Information of task completion.
+     */
+    private boolean completed;
```
bryanljx: Might be more appropriate here to name `completed` as `isCompleted` to be consistent with the coding standards defined for CS2103T though there are certainly cases where I would prefer `completed` to `isCompleted`

src/main/java/duke/command/DeadlineCommand.java
```diff
@@ -0,0 +1,105 @@
...
+     * @param t Deadline task.
+     * @param date Date deadline for task.
+     * @throws DukeException
+     */
+    private static void dateSetter(Task t, String date) throws DukeException{
+        String[] dateTime = date.split(" ", 2);
+
+        String ddmmyyyyRegex = "[0-9]{1,2}/[0-9]{1,2}/[0-9]{4}";
+        String yyyymmddRegex = "[0-9]{4}-[0-9]{1,2}-[0-9]{1,2}";
+        String timeRegex = "[0-9]{4}";
```
bryanljx: Sick! Nice use of regular expressions here 👍

bryanljx (commented):
Overall well written and compliant with the coding standard set in CS2103T. 

However, there are a few areas where there are redundant empty lines inserted which looked quite weird and spacing and line breaks could be more consistent as well. These are quite common mistakes that we will make so I would suggest actually setting up a code formatter or IntelliJ's auto format on save.

Keep up the good work!

PR #256, opened 2022-08-23
[Poomklao] iP

parnikkapore:
*Note: This PR is opened as part of course requirements.*

# y

> [the PR] provides a convenient way for us to access the current state of all your iP code from one location.

1. Training
1. Workload padding (not)
1. Because It's Fun!™ 😁

# Try it out

Builds are provided for selected revisions in the [Releases section](https://github.com/parnikkapore/ip/releases).

# Regression testing

To run the regression tests, run the following from the clone's root directory:

```bash
cd text-ui-test
./runtest.sh
```

Support for regression testing under Windows is currently broken.

# Todo

* [ ] Fix regression testing on Windows (`runtest.bat` needs an update)
* [ ] Make a GUI

# Misc notes

* 90s mascot energy (well that's what Duke was, isn't he)
* Plugin system was originally for completely different and decoupled functionalities


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,42 @@
...
+
+        // initialize plugins
+        List<String> messages = new ArrayList<>();
+
+        Todos todos = new Todos();
+        messages.addAll(todos.init());
+
+        printLines(messages);
+
+        boolean fExit = false;
```
jasonqiu212: does `fExit` mean forceExit? Perhaps a more intuitive boolean variable name here?

src/main/java/duke/ExecuteResult.java
```diff
@@ -0,0 +1,37 @@
...
+public class ExecuteResult {
+    /**
+     * Whether the program should exit after this command
+     */
+    private final boolean shouldExitAfter;
+
+    /**
+     * Lines of text to be sent as a reply from this command
+     */
+    private final List<String> reply;
```
jasonqiu212: would it be better to name this in plural form, since it's a collection of objects?

src/main/java/duke/FallibleFunction.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Represents a function that can either complete and return a value or throw a MessagefulException for the
+ * caller to handle.
+ *
+ * @param <T> the type of the input to the function.
+ * @param <R> the type of the input to the function.
+ */
+@FunctionalInterface
+public interface FallibleFunction<T, R> {
```
jasonqiu212, line 11: I really like this implementation of functional programming!

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Format an error message.
+     *
+     * @param e The MessagefulException to output.
+     */
+    public static List<String> sayError(MessagefulException e) {
+        if (DEV_MODE) {
+            return sayAsError(e.toString());
+        } else {
+            return sayAsError(e.getHint());
```
jasonqiu212, line 50: i like that you have a dev_mode

src/main/java/duke/modules/todos/Event.java
```diff
@@ -0,0 +1,98 @@
...
+            return new Event(match.group("name"), match.group("timeRange"));
+        } else {
+            throw new MessagefulException(
+                    "Event syntax no match",
+                    "Events are added like this: event project meeting /at Mon 2-4pm"
+            );
+        }
+    }
+
+    public static final String typeCode = "E";
```
jasonqiu212: i think the `typeCode` should be in all caps separated by underscores, since it's a constant. A comment here would be good to explain the purpose of this constant

src/main/java/duke/gui/MainWindow.java
```diff
@@ -0,0 +1,79 @@
...
+    @FXML
+    private ScrollPane scrollPane;
+    @FXML
+    private VBox dialogContainer;
+    @FXML
+    private TextField userInput;
+    @FXML
+    private Button sendButton;
+
+    private Image userImage = new Image(this.getClass().getResourceAsStream("/media/Damith.png"));
```
jasonqiu212, line 30: nice hahaha

jasonqiu212 (commented):
hello! your code looks great to me! I especially like the usage of functional programming and dev_mode! just some minor coding standard issues here and there 👍

src/main/java/duke/Duke.java
parnikkapore: I think it stood for `flag_Exit`, but yeah, this should be renamed `willExit`.

src/main/java/duke/ExecuteResult.java
parnikkapore: You do have a point, as it's not treated as a single object all the way down.

src/main/java/duke/modules/todos/Event.java
parnikkapore: You have a point. I agree about needing the comment.

PR #257, opened 2022-08-23
[Ho Lok Wah] iP

ThomasHoooo:
- text-based
- easy to learn

All you need to do is,
2. double-click it.
3. add your tasks.

And it is **FREE**!
Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


gabyang (commented):
Wow what a great application!

ThomasHoooo:
> Wow what a great application!

Thank you for being honest!

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,36 @@
...
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    /**
+     * Unmarks the task as not done.
+     */
+    public void unmark() {
+        this.isDone = false;
+    }
```
donljh, line 25: Perhaps you could consider naming these two methods similarly since they perform similar functions? 

i.e. Renaming unmark to markAsUndone, or markAsDone to simply mark

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,47 @@
...
+        tasks.get(index).markAsDone();
+    }
+
+    public void unmark(int index) {
+        tasks.get(index).unmark();
+    }
+
+    public TaskList find(String task) {
+        return new TaskList((ArrayList) tasks.stream().filter(s -> s.toString().contains(task)).collect(Collectors.<Task>toList()));
+    }
```
donljh, line 45: I see that this method tries to find a keyword that each task might contain in its description. Perhaps it might be clearer to name the relevant method parameter that way?

i.e. Rename task to keyword in the method parameter

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,82 @@
...
+                Ui.printDeletedMessage(deletedTask, tasks.getSize());
+                storage.writeToFile(tasks);
+                continue;
+            }
+            case MARK:
+                tasks.markDone(Integer.parseInt(command.replace("mark ", "").trim()));
+                storage.writeToFile(tasks);
+                continue;
+            case UNMARK:
+                tasks.unmark(Integer.parseInt(command.replace("mark ", "").trim()));
```
donljh: Not sure if this was a mistake, but was your intention to replace the "unmark" word instead?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,82 @@
...
+                Task deletedTask = tasks.remove(deleteItem);
+                Ui.printDeletedMessage(deletedTask, tasks.getSize());
+                storage.writeToFile(tasks);
+                continue;
+            }
+            case MARK:
+                tasks.markDone(Integer.parseInt(command.replace("mark ", "").trim()));
+                storage.writeToFile(tasks);
+                continue;
+            case UNMARK:
```
donljh: I like that you used the switch clause to effectively separate the logic of each command. However, some commands seem to require parsing the user input in similar fashions (mark and unmark, deadline and event). 

Perhaps it might be better to abstract the parsing, as well as each command into their own classes to better utilize OOP?

src/main/java/duke/Task.java
ThomasHoooo: Don't think this is possible because I have 2 commands, mark and unmark. If I combine the two methods I won't be able to check whether it's done or not done.

src/main/java/duke/Duke.java
ThomasHoooo: Good point

PR #260, opened 2022-08-23
[Yeo Jun Jie] iP

yeojunjie:
# Duke

## Dukescription

Duke is able to do some *things* to help you manage your schoolwork. Some of the *things* include:
* adding tasks
* deleting tasks ~to escape from them~
* completing tasks
* searching for tasks

There are many types of ~pain~ tasks. Duke supports
1. Events
2. Deadlines
3. To-dos

## Features
- [X] Manage tasks
- [ ] Order tasks by deadline (coming soon)
- [ ] A help menu which appears when the user types `help` or when the user input is, let's just say, open to interpretation. (coming soon)

## Cancelled Features

- [ ] ~A personality. In contrast to [Sir Duke](https://www.youtube.com/watch?v=6sIjSNTS7Fs), this program will take on the personality of an Encik. We call it an Encik Duke.~
This idea was rejected because it uses a very specialised vocabulary, which would not be appropriate for international audiences. 🫤 An example would be the following phrase.
    > one times good one

```
// This is a comment.
/* This is also a comment. */
```


data/duke.txt
```diff
@@ -0,0 +1,9 @@
+T|1|read book
+D|1|return book|2022-02-02T22:22
```
boredcoco, line 8: Perhaps this file should be deleted before committing so that the user can test the code to see if it works without a data/duke.txt file being generated at first?

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,29 @@
+package duke;
+
```
boredcoco, line 2: Consider creating a separate package for all classes that extend Task within the Duke package.

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    @Test
+    public void execute_validEventAddition_eventIsAdded() {
+        TaskList taskList = new TaskList();
+        Parser parser = new Parser(taskList);
+        parser.execute("event New Year's Day 2023 @ 2023-01-01T00:00:00");
+        assertEquals(taskList.size(), 1);
+    }
+}
```
boredcoco, line 26: Perhaps you can add more test cases?

src/main/java/duke/CommandType.java
```diff
@@ -0,0 +1,12 @@
...
+public enum CommandType {
+    EXIT,
+    ADD,
+    REMOVE,
+    MARK_AS_DONE,
+    MARK_AS_UNDONE,
+    SEARCH,
+    LIST,
+    NONSENSE
+}
```
boredcoco: I like the use of enums here.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,52 @@
...
+import duke.Parser;
+
+/**
+ * Represents an instance of Duke.
+ */
+public class Duke {
+
+    private Storage storage;
+    private TaskList taskList;
+    private Ui ui;
```
boredcoco: I like how you extracted out the UI and various components, good job!

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,29 @@
+package duke;
```
boredcoco, line 1: Like Deadline, you can consider putting this in a separate package within the Duke package so that it is more organised.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,128 @@
...
+        } else if (userCommand.startsWith("unmark")) {
+            return CommandType.MARK_AS_UNDONE;
+        } else if (userCommand.startsWith("list")) {
+            return CommandType.LIST;
+        } else if (userCommand.startsWith("search")) {
+            return CommandType.SEARCH;
+        } else {
+            return CommandType.NONSENSE;
+        }
+    }
```
boredcoco: I like the use of enums here, good job!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,107 @@
...
+import duke.ToDo;
+import duke.Deadline;
+import duke.Event;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.Objects;
+import java.util.Scanner;
```
boredcoco, line 13: I like the way you organised your imports so it is neater and easier to read, good job!

boredcoco (commented):
Overall, I like the way you tried to incorporate SWE principles into IP. Although at this point of the review you have yet to start on Level-10, thus far your project looks great! You could incorporate more test cases and put all classes that inherit Task into a separate package for better organisation. I hope the suggestions given are useful.

PR #261, opened 2022-08-23
[Gerald Teo] iP

jialatteo:

- text-based
- easy to learn
- ~FAST~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/jialatteo/ip/releases/tag/v0.1)
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon) 

If you are a Java programmer, you can use it too practice Java too. 
Here's the `main` method:
```java
        new Duke("tasks.txt").run();
    }

```


src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,116 @@
...
+     * @return List This is the current list of the user's tasks
+     */
+    public List<Task> load() {
+        File file = this.getFile();
+        List<Task> taskList = new ArrayList<>();
+        try {
+            Scanner sc = new Scanner(file);
+            while (sc.hasNextLine()) {
+                String line = sc.nextLine();
+                String[] splitStr = line.split("\\|");
```
clarence-chew, line 64: The way this is implemented seems to inconvenience users wanting to use the `|` symbol in their task description.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+     * This method takes in the users string input
+     * and returns a list of tasks that contains
+     * the user's search keyword
+     * @param str Command input from user
+     * @param taskList User's complete task list
+     * @return list of tasks containing user's search keyword
+     * @throws DukeException This tells the user if no such task contains the user's search keyword
+     */
+    private static List<Task> getAllTaskWithKeyword(String str, TaskList taskList) throws DukeException {
+        String[] splitStr = str.split(" ");
```
clarence-chew: There's a bit of code duplication here where the idea of "splitting and matching the first word, then getting the rest of the string" is repeated multiple times. Can this common idea be perhaps abstracted away, perhaps with other functions like this?
```java
public static boolean isOfTheCommand(String input, String commandName) {
    String[] splitStr = input.split(" ");
    return splitStr[0].equals(commandName));
}
public static boolean getCommandBody(String input) {
    String[] splitStr = input.split(" ");
    String type = splitStr[0];
    return input.substring(type.length() + 1);
}
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Parser class allows the Duke program
+ * to parse user inputs into commands.
+ *
+ * @author Gerald Teo Jin Wei
+ * @version 0.1
+ * @since 2022-08-28
+ */
+public class Parser {
+    public Parser() {}
```
clarence-chew: This seems to be dead code. One way to perhaps make it somewhat useful would be to write it as
`private Parser() {}`, to make it such that the class cannot be instantiated.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,43 @@
...
+        super(description);
+        this.at = date;
+    }
+
+    /**
+     * Returns the task's type, completion status, description and date
+     * @return String This returns the string of the event in the specified format
+     */
+    public String toString() {
+        return "[E][" + this.getStatusIcon() + "] " + this.description + " (at: " + at + ")";
```
clarence-chew: Do avoid unnecessary use of `this`, as per the coding standard given.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+                } else if (taskToAdd != null) {
+                    ui.addTask(taskList, taskToAdd);
+                } else if (taskToDelete != null) {
+                    ui.deleteTask(taskList, taskToDelete);
+                } else if (taskListWithKeyWord.size() != 0) {
+                    ui.printTasksWithKeyword(taskListWithKeyWord);
+                } else {
+                    throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(\n");
+                }
+            }
```
clarence-chew: The indentation in this area may be reduced through the use of guard clauses, i.e.
```java
if (command.equals(bye) {
    ui.quit();
    return;
}
if (command.equals(list) {
    ...
    return;
}
String[] splitStr = ...
...
if (taskToMark != null) {
    ...
    return;
}
if (taskToAdd != null) {
    ...
    return;
}
...
throw new DukeException("...");
```

clarence-chew (commented):
Nice work with your chatbot!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+ * @since   2022-08-28
+ */
+public class Deadline extends Task {
+    protected String by;
+
+    /**
+     * Constructor for Deadline class
+     * @param description Description of deadline
+     * @param date Due date of deadline
+     * @param isDone Completion status of deadline
```
midnightfeverrr, line 18: Comments should have empty line between description and parameter section

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,43 @@
...
+    public Deadline(String description, String date) {
+        super(description);
+        this.by = date;
+    }
+
+    /**
+     * Returns the task's type, completion status, description and due date
+     * @return String This returns the string of the deadline in the specified format
+     */
+    public String toString() {
```
midnightfeverrr, line 40: Missing @override tag

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
+
+    /**
+     * This method takes in the string input of the user
+     * and return the task from the taskList to be marked/unmarked.
+     * When an invalid number is entered, user will get to
+     * enter again
```
midnightfeverrr: Do consider rephrasing comments according to match java coding standard (i.e. MARKS the task from taskList to be done/undone.)

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,116 @@
...
+ * all its tasks and their respective states in a txt file.
+ * The txt file will be created if it does not exist in the filePath
+ *
+ * @author Gerald Teo Jin Wei
+ * @version 0.1
+ * @since 2022-08-28
+ */
+public class Storage {
+
+    private String filePath;
```
midnightfeverrr: consider making the filePath final

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+            boolean validAction = action.equals("delete");
+            boolean validIndex = index > 0 && index <= taskList.getSize();
+            if (validIndex && validAction) {
+                return taskList.getTask(index - 1);
+            } else if (validAction && !validIndex) {
+                throw new DukeException("☹ OOPS!!! Please enter a valid task number to delete");
+            }
+            return null;
+        } catch (NumberFormatException e) {
+            return null;
```
cfyjoanna: We should avoid empty catch statements. Maybe you can add a comment explaining why it's okay to return null here?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+            if (validIndex && validAction) {
+                return taskList.getTask(index - 1);
+            } else if (validAction && !validIndex) {
+                throw new DukeException("☹ OOPS!!! Please enter a valid task number to mark/unmark");
+            }
+            return null;
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
```
cfyjoanna: The code here looks very similar to the one in `getTaskToDelete`. To improve readability, perhaps you could write a separate method to process the checking of valid index?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,202 @@
...
+                throw new DukeException("☹ OOPS!!! Please set date of deadline with /by.\n");
+            }
+            String description = str.substring(type.length() + 1, str.indexOf("/by") - 1);
+            String date = str.substring(str.indexOf("/by") + 4);
+            String formattedDate = Parser.parseDate(date);
+            return new Deadline(description, formattedDate);
+        } else if (type.equals("event")) {
+            if (splitStr.length < 2) {
+                throw new DukeException("☹ OOPS!!! The description of an event cannot be empty.\n");
+            }
```
cfyjoanna: Maybe you could perform the checking of splitStr.length < 2 before checking for the type of task to avoid repeated checking and reduce the length of code. The exception message could be changed to something more generic.

cfyjoanna (commented):
There are only some minor changes that can be implemented to improve readability. Looks like your code is great otherwise. Nice job! :)

PR #262, opened 2022-08-23
[Min Hein Aung] iP

MinHeinA:
### Duke

> "You can do anything, but not everything." – David Allen

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/MinHeinA/ip).
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's the main method:

```
    Duke duke = new Duke("data/duke.txt");
    duke.run();
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,63 @@
+package duke;
+
+import duke.command.Command;
+
+public class Duke {
+    public Ui ui;
+    public TaskList taskList;
+    public Storage storage;
```
lkwlkww: I'm not sure if these variables are accessed by other classes within your implementation - if not, do you think the access modifier could be changed to `private`?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+            if (file.exists()) {
+                try {
+                    Scanner scanner = new Scanner(file);
+                    while (scanner.hasNextLine()) {
+                        String nextLine = scanner.nextLine();
+                        String[] splitString = nextLine.split("\\|");
+
+                        Task task;
+                        switch (splitString[0]) {
```
lkwlkww: Perhaps you could unindent the `case` lines here? According to the coding standard [guidelines](https://se-education.org/guides/conventions/java/index.html#variables):
> Note there is no indentation for case clauses.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,49 @@
+package duke;
+
+import java.util.Scanner;
+
+public class Ui {
+    public static final String DIVIDER = "____________________________________________________________";
```
lkwlkww: For my dividers, I used the `repeat` method of a String, which I thought would make it look cleaner. For example:
```java
"_".repeat(5) // Will become _____
```

My apologies for assuming you didn't know about this method. If you did, could you share why you chose not to do it that way?

lkwlkww (commented):
Overall code quality seems really good!

tanhl2000:
Overall code quality is good, coding standard is followed, other than the indentation of cases that was previously pointed out.

tanhl2000:
I actually like that you did not use the String.repeat() function as i think it improves readability of the code. However definitely understand that it is a QOL improvement to use the repeat() function and i also did not know about the repeat function before

tanhl2000:
for the DukeException however i believe it would be better if you could create individual exceptions for each exception case as it would make the exception handling structure better! right now a DukeException being thrown will not let us know anything unless we read the message that is also thrown, but i think the exception itself should give some information as well

tanhl2000:
but overall really good work! :)

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,21 @@
+package duke;
+
+/**
+ * A dedicated Exception class for Duke program.
+ *
+ */
+public class DukeException extends Exception {
```
tanhl2000: for the DukeException however i believe it would be better if you could create individual exceptions for each exception case as it would make the exception handling structure better! right now a DukeException being thrown will not let us know anything unless we read the message that is also thrown, but i think the exception itself should give some information as well (sorry for the repeat comment I had not put it in the PR review section properly)

src/main/java/duke/Ui.java
tanhl2000: I actually like that you did not use the String.repeat() function as i think it improves readability of the code. However definitely understand that it is a QOL improvement to use the repeat() function and i also did not know about the repeat function before!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,78 @@
+package duke;
+
+import duke.command.Command;
```
tanhl2000: Overall code quality is really good and coding standards are followed (other than the indentation fault in the case-switch section that was previously pointed out.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,78 @@
...
+                command.execute(taskList, storage, ui);
+                if (command.isExit()) {
+                    break;
+                }
+            } catch (DukeException exception) {
+                ui.printMessage(exception.toString() + "\n");
+            }
+        }
+    }
+     */
```
anuanas2007: Perhaps you could decide whether to delete it out or not?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,46 @@
...
+            return new AddCommand(userInput);
+        } else if (command.equals(Duke.Keyword.DEADLINE.getKeyword())) {
+            return new AddCommand(userInput);
+        } else if (command.equals(Duke.Keyword.DELETE.getKeyword())) {
+            return new DeleteCommand(userInput);
+        } else if (command.equals(Duke.Keyword.FIND.getKeyword())) {
+            return new FindCommand(userInput);
+        }else {
+            throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
```
anuanas2007: Perhaps using a switch statement might be an easier way to implement it?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,78 @@
+package duke;
+
```
anuanas2007: Maybe you can consider changing your bot name?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+/** Storage object used for file operations such as read and write of TaskList to text file. */
+public class Storage {
+
+    private Ui ui;
```
anuanas2007: Storage class looks good!!

src/main/java/duke/command/MarkCommand.java
```diff
@@ -0,0 +1,45 @@
...
+import duke.Ui;
+import duke.task.Task;
+
+/**
+ * Represents the command to mark specific Tasks in TaskList as completed or not completed
+ * that inherits from Command.
+ */
+public class MarkCommand extends Command {
+    private final String userInput;
+    private final boolean isMark;
```
anuanas2007: Maybe you consider changing the name of the class to better represent both marking and unmarking command.

anuanas2007 (commented):
Overall, your code looks good. Just a few suggestions you can consider!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,119 @@
...
+    private Ui ui;
+
+    /**
+     * Represents a TaskList object.
+     * Initialises a new empty ArrayList of Tasks and Initialises ui to print responses when task methods are executed
+     *
+     */
+    public TaskList() {
+        this.taskList = new ArrayList<>();
+        this.ui = new Ui();
```
clarence-chew: Coding standard recommends avoiding unnecessary use of `this` with fields.
https://se-education.org/guides/conventions/java/index.html#variables

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,78 @@
...
+                ui.printMessage(exception.toString() + "\n");
+            }
+        }
+    }
+     */
+
+    public String getResponse(String userInput) {
+        try {
+            Command command = Parser.parse(userInput);
+            return command.execute(taskList, storage, ui);
```
clarence-chew: I like how you made your Command class such that it is executed once at an obvious place.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,133 @@
...
+                            break;
+                        }
+                        case "E": {
+                            task = new Event(splitString[2], splitString[3]);
+                            break;
+                        }
+                        default:
+                            throw new DukeException("OOPS!!! Invalid task type found in file!");
+                        }
+
```
clarence-chew: Perhaps some sort of Factory design pattern could be useful here?

PR #268, opened 2022-08-23
[Zhaoqi] iP

Eclipse-Dominator:
# Duke 

A magical checklist application

------

## what does duke have to offer

- Magical yet mysterious wonders

- ~~Hard~~  **Easy**  to use

- Tonnes of features

  ------

  

## what are you waiting for?

1. Download the [future](https://www.youtube.com/watch?v=dQw4w9WgXcQ) now!
2. Please don't click on the link above, you *will* regret it.

>  "Yup, best of all, it's **Ad-free**"
>
> "Wonderful, what brilliancy!"

​				~Users

Features:

- [ ] CLI pattern highlighting (only works on cli and not gui sadly)
- [x] Add Todo Task
- [x] Add Deadline Task
- [x] Add Event Task
- [x] Automatic datetime parsing if input format is datetime

------

There is even some python code to generate Test files too!

```Python
sample_header = '''
import org.junit.jupiter.api.Test;
import static org.junit.jupiter.api.Assertions.assertEquals;
public class {} {{
}}
'''

source_folder = "./src/main/java/"
test_folder = "./src/test/java/"


for root, dirs, files in os.walk(source_folder):
    curr_path_in_test = f"{test_folder}{root[len(source_folder):]}"
    print(f"generating dir at {curr_path_in_test}")
    for dir in dirs:
        curr_path = os.path.join(curr_path_in_test, dir)
        if not os.path.isdir(curr_path):
            os.mkdir(curr_path)
     # ....
```


src/main/java/duke/main/Duke.java
```diff
@@ -0,0 +1,116 @@
...
+            + "|  _ \\ _   _| | _____ \n"
+            + "| | | | | | | |/ / _ \\\n"
+            + "| |_| | |_| |   <  __/\n"
+            + "|____/ \\__,_|_|\\_\\___|\n"
+            + "      Chatbot!\n";
+
+    private static final String INTRO = "Hello! I'm Duke\n"
+            + "What can I do for you?";
+
+    private DukeIo userIO;
```
dexter-sim: The textbook states "Abbreviations and acronyms should not be uppercase when used as a part of a name.".
Do you think it is better to name it `userIo` or `userInputOutput`?

src/main/java/duke/main/Duke.java
```diff
@@ -0,0 +1,116 @@
...
+            userIO.printError(e);
+            userIO.printTask("Fatal Error! The system will exit abnormally!");
+            return null;
+        }
+
+        return new Duke(tasks, dukeData);
+    }
+
+    /**
+     * Initiate the running loop. Will exit when <code>handleInput</code> returns false.
```
dexter-sim: The textbook states "In method header comments, the first sentence should start in the form `Returns ...`, `Sends ...`, `Adds ...` (not `Return` or `Returnning` etc.)".
Should this line start with `Initiates` instead?

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param data ParsedData containing information for both
+     * @return Deadline
+     * @throws DukeException Throws when data given is invalid/insufficient
+     */
+    public static Deadline createDeadline(ParsedData data) throws DukeException {
+        if (data.description.length() == 0)
+            throw new EmptyDescriptionException("deadline");
+
+        if (data.additionalInfo.length() == 0)
+            throw new EmptyTimeException("deadline", SPLIT);
```
dexter-sim: The textbook states "Single statement conditionals should still be wrapped by curly brackets." and "Use K&R style brackets (aka [Egyptian style](https://blog.codinghorror.com/new-programming-jargon/)).".
Would you consider adding curly braces for these 2 single statement conditionals?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Optional;
+
+import duke.util.ParsedData;
+
+/**
+ * Parent task class for all other Tasks.
+ */
+public abstract class Task implements Comparable<Task> {
+    protected final String description;
+    protected boolean completed;
```
dexter-sim: The textbook states "Boolean variables/methods should be named to sound like booleans.". 
Do you think `hasComplete` or `isComplete` would be more suitable name for a boolean variable?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,97 @@
...
+    public String toString() {
+        return String.format("[%s] %s", (completed) ? "X" : " ", description);
+    }
+
+    /**
+     * Convert the current task data into parsedata to be saved.
+     * 
+     * @return converted Task
+     */
+    abstract public ParsedData convertToParseData();
```
dexter-sim: The textbook states "Method modifiers should be given in the following order: 
`<access> static abstract synchronized <unusual> final native`".
Would you agree that it should be ```public abstract ParsedData convertToParseData();``` instead?

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,50 @@
...
+                Matcher.quoteReplacement(searchString));
+
+        List<Task> tasks = taskList.getTasks();
+
+        io.printLine();
+        io.printTask(FIND_TXT, 0);
+        String sBefore, sAfter;
+        for (int i = 0; i < tasks.size(); i++) {
+            sBefore = tasks.get(i).toString();
+            sAfter = sBefore.replaceAll(searchString, RED_TXT);
```
dexter-sim: The textbook states "Variables should be initialized where they are declared and they should be declared in the smallest scope possible."
Do you think that the declaration of the 2 strings be in the for loop instead of before the for loop?

dexter-sim (commented):
Looks good.
Some minor coding standard violations.

src/main/java/duke/command/ByCommand.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class ByCommand extends DataCommand {
+
+    /**
+     * Takes in ParsedData potentially containing a datetime pattern.
+     * 
+     * @param d ParsedData containing a possible datetime pattern
+     */
+    public ByCommand(ParsedData d) {
+        super(d);
```
ElijahS67: Perhaps a more descriptive name rather than "d" is better?

src/main/java/duke/command/TodoCommand.java
```diff
@@ -0,0 +1,46 @@
...
+import duke.util.TaskList;
+
+/**
+ * Command to add a todo task to the tasklist.
+ */
+public class TodoCommand extends DataCommand {
+
+    private static final String ADD_TASK = "Got it. I've added this task:%n"
+            + "  %s%n"
+            + "Now you have %d tasks in the list.";
```
ElijahS67: Perhaps the indentation of line 20 should be 8 spaces more than the parent line?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * {@inheritDoc} Compares order based on data assigned to task.
+     * 
+     * @param o The other task
+     * @return a task is bigger if its time higher and biggest if non-existent
+     */
+    @Override
+    public int compareTo(Task o) {
```
ElijahS67: Perhaps a more descriptive name than "o" should be used?

ElijahS67 (commented):
Looks good to merge!

src/main/java/duke/main/Duke.java
Eclipse-Dominator: Updated.

src/main/java/duke/main/Duke.java
Eclipse-Dominator: updated.

src/main/java/duke/task/Deadline.java
Eclipse-Dominator: definitely, I had made the changes for this one but forgot to push it.

src/main/java/duke/task/Task.java
Eclipse-Dominator: updated to fit coding standard.

src/main/java/duke/task/Task.java
Eclipse-Dominator: definitely, I had made the changes for this one but forgot to push it.

src/main/java/duke/command/FindCommand.java
Eclipse-Dominator: Yup, moved the declaration to within the for loop.

src/main/java/duke/command/ByCommand.java
Eclipse-Dominator: changed d -> parsedData.
Although for under the scope of the constructor which is really short, I think this is ok 

>  Variables with a large scope should have long names, variables with a small scope can have short names.

src/main/java/duke/command/TodoCommand.java
Eclipse-Dominator: Line 20's parent line should still be line 18, so I don't think there should another 8 spaces on line 20. 
I think the parent line should change when there is a change in scope. 
ie 
```
demoSomeRandomMethodThatIsLong()
        .method1()
        .method2(demo2        // <-- no change in parent line
                .method3());        // <-- change in parent line
```

src/main/java/duke/task/Task.java
Eclipse-Dominator: changed `o` -> `task`.
Although for under the scope which is short, I think this keeping it as o is fine

>  Variables with a large scope should have long names, variables with a small scope can have short names.

PR #274, opened 2022-08-23
[Lee I-Shiuan] iP

ish1506:
# Duke 

>Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))

Duke helps you stay on top of your work. It's:

* text-based
* easy to learn
* ~~FAST~~ *SUPER FAST* to use 😃

All you need to do is:

1. download it from [here](https://github.com/ish1506/ip/releases/tag/A-Jar).
2. use the `java -jar duke.jar` command to run the program.
3. add your tasks. 
4. let it manage your tasks for you!

And it's **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

---

If you're a Java programmer, you can use it to practice Java too. Here's the `main` method:

```java
public class Duke {
        new Duke().run();
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,97 @@
...
+import duke.parser.Parser;
+import duke.storage.Storage;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.TaskList;
+import duke.task.Todo;
+import duke.ui.Ui;
+
+import java.util.Scanner;
```
JasonCP14: Perhaps a better way to write import statements is to order them on:
1. Java Packages, then Custom Packages
2. Lexical Order

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Parses the user input to get the command and its arguments.
+     *
+     * @throws DukeException if the input cannot be parsed.
+     */
+    public void parse() throws DukeException {
+        List<String> listOfInputs = Arrays.asList(input.split(" "));
+        if (input.isEmpty() || listOfInputs.isEmpty()) {
```
JasonCP14, line 48: Don't you think a series of switch case statements will be better than such large number of if else statements?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,97 @@
+package duke;
```
JasonCP14, line 1: Your variable names are intuitive and are following the coding standard. Nicely done!

JasonCP14 (changes requested):
LGTM, almost. Just a few nits to fix.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,97 @@
...
+            try {
+                Parser parser = new Parser(input);
+                parser.parse();
+                Command command = parser.getCommand();
+                int taskIndex = parser.getTaskIndex();
+                String[] args = parser.getArgs();
+
+                switch (command) {
+                case BYE:
+                    // Write tasks data to storage before terminating program
```
isanidiot: Perhaps you can do this after each time the list is updated?

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,5 @@
+package duke.command;
+
```
isanidiot, line 2: I like how the enum has its own class.

isanidiot (commented):
Keep up the good work in week 5! I know GUI is hard...

PR #285, opened 2022-08-23
[Hao Sitong] iP

o-ohst:
# Karen Bot

**User review:**
> ⭐️⭐️⭐️⭐️⭐️ "This bot is amazing! Since using Karen Bot, I no longer have any todos or deadlines, cuz I no longer have a job!" -Damith

**Features:**
- simple
- lightweight
- unnecessary

How to use:

1. Go to [this link](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

Instructions to get started:

- [ ] Delete this app
- [ ] Don't come back again

Run this `main` method to get started:
```java
public static void main (String[] args) {
    System.out.println("Hello world");
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,30 @@
+package duke;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
+    protected LocalDateTime by;
```
NicholasTYD, line 7:
```suggestion
    protected LocalDateTime dueDate;
```
Perhaps it would be better if you use dueDate instead of by? That could make field sound more intuitive especially when its referenced in other parts of the code.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,42 @@
...
+    
+    public Task get(int index) {
+        return taskList.get(index);
+    }
+
+    public ArrayList<Task> getTaskList() {
+        return this.taskList;
+    }
+    
+    public int size() {
```
NicholasTYD, line 31:
```suggestion
    public int getTaskCount() {
```
Would it be better to rename this method to sound more like a verb? It may be more easily understood when calling this method from another class.

src/main/java/duke/Todo.java
```diff
@@ -0,0 +1,24 @@
+package duke;
+
+public class Todo extends Task{
+    protected String by;
```
NicholasTYD: Is this field used? It might be better to remove it if its not needed to reduce unnecessary code.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,66 @@
...
+    public void sendFailureMessage() {
+        System.out.println(divider);
+        System.out.println("What's that??");
+        System.out.println(divider);
+    }
+    
+    public void sendEmptyTaskMessage() {
+        System.out.println(divider);
+        System.out.println("Empty task? Are you kidding me??");
+        System.out.println(divider);
```
NicholasTYD:
```suggestion
        String s = divider + "\n" 
                + "Empty task? Are you kidding me??" + "\n" 
                + divider;
        System.out.println(s);
```
Would it be better to combine all the messages into one string before printing it out? It may reduce the need to call System.out.println() many times.

NicholasTYD (commented):
Overall, I found your code easy to read for the most part except a few places
where you could use more meaningful names for your fields. I noted a few areas where you could write more concise code as well. Otherwise, your iP looks pretty great so far!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,157 @@
...
+                ui.sendByeMessage();
+                break;
+            }
+            if (text.equals("list")) {
+                ui.sendList(tasks);
+                continue;
+            }
+            if (text.startsWith("mark ")) {
+                int index = Integer.parseInt(text.split(" ")[1]) - 1;
+                if (index < 0 || index >= tasks.size()) {
```
xiaobill8: Hi Sitong, you could do this in a try catch block instead since we are dealing with ArrayIndexOutOfBoundsException here

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,49 @@
...
+     * @return LocalDateTime
+     */
+    public static LocalDateTime parseDateTime(String s) {
+        String[] dtStrings = s.split(" ");
+        String date = dtStrings[0];
+
+        String[] dateArr = date.split("/");
+        String time = dtStrings[1];
+
+        LocalDateTime dt = LocalDateTime.of(
```
xiaobill8, line 38: Perhaps you could use the DateTimeFormatterBuilder to create a custom Datetime format to accept in parsing, rather than constructing your own datetime

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,32 @@
+package duke;
+
+import java.time.LocalDateTime;
+
+public abstract class Task {
```
xiaobill8, line 5: I like how you encapsulated task into an abstract class

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,157 @@
...
+        storage.save(tasks);
+    }
+    
+    public void run() {
+        ui.sendStartMessage();
+        Scanner input = new Scanner(System.in);
+        while (true) {
+            String text = input.nextLine();
+            if (text.equals("bye")) {
+                ui.sendByeMessage();
```
xiaobill8, line 63: Perhaps you could encapsulate each of these commands into a Command object so that less of the logic is present in the main Duke.java code

xiaobill8 (commented):
Very concise and well-written code overall!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,49 @@
...
+                Integer.parseInt(dateArr[2]),
```
xiaobill8, line 39: Basic indentation should be 4 spaces instead of 8 here

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,79 @@
...
+    
+    private final File file;
+    
+    public Storage(String filePath) {
+        file = new File(filePath);
+    }
+
+    /**
+     * Load tasks from storage or create new file.
+     * @return ArrayList of tasks
```
xiaobill8, line 20: There should be an empty line between description and parameter/return section

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,79 @@
...
+    public ArrayList<Task> load() {
+        if (file.exists()) {
+            return loadFromFile();
+        } else {
+            return createFile();
+        }
+    }
+
+    /**
+     * Save all tasks to storage
```
xiaobill8, line 31: There should be an empty line between description and parameter/return section

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,49 @@
+package duke;
+
+import java.time.LocalDateTime;
+
+public class Parser {
+
+    /**
+     * Decodes a task from the given encoded string
```
xiaobill8, line 8: There should be an empty line between description and parameter/return section

src/test/java/duke/TodoTest.java
```diff
@@ -0,0 +1,13 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TodoTest {
+    @Test
+    public void newTodoCreated(){
```
xiaobill8, line 9: Perhaps you could name this as newTodoCreationTest()

xiaobill8 (commented):
Sorry, figured that the review should be on coding standards. Generally well done! Just take note of indentation and empty line between JavaDoc description and param/return section.

PR #286, opened 2022-08-23
[Damien Koh] iP

Ssatu:
##Duke
> Duke duke duke duke
Who is [Duke](https://www.oracle.com/java/duke/)?
~~Why is Duke?~~

What can Duke do for _you_?
* Manage your to-do list
* Keep you informed of your deadlines
* Help you find information about your tasks with ease

Where can you use Duke? 👍 
1. On the go,
2. at home
3. basically **anywhere**!

What are some of the tasks that you can schedule?
- [x] To-do
- [x] Deadlines
- [x] Events
- [ ] Reminders

If you're a java programmer you can use it to practice Java too! Here's the `main` method:
```java
        new Duke("data.txt").run();
    }
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,84 @@
...
+public class Parser {
+
+    /**
+     * Parses a user's input, and calls the necessary functions.
+     *
+     * @param input The user's input.
+     * @param tasks The TaskList being used.
+     * @param storage The storage object being used.
+     */
+    public static String parse(String input, TaskList tasks, Storage storage) {
```
Tex-Tang, line 19: Should the `parse` method do only parsing command instead of calling the command?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,84 @@
...
+        } else if (first.equals("find")) {
+            try {
+                return tasks.find(rest);
+            } catch (DukeException ex) {
+                return ex.toString();
+            }
+        } else {
+            return "OOPS!!! I'm sorry, but I don't know what that means :-(";
+        }
+    }
```
Tex-Tang: Should we wrap these code into single `try-catch` to avoid repeating `try-catch` for each command.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,225 @@
...
+        int found = 0;
+        for (int i = 0; i < tasks.size(); i++) {
+            if (tasks.get(i).getTitle().contains(input)) {
+                found += 1;
+                //Add index
+                list += (found) + ".";
+                //Add task
+                list += tasks.get(i);
+                //Add nextline
+                list += "\n";
```
Tex-Tang, line 227:
```suggestion
                list += String.printf("%d. %s\n", found, tasks.get(i));
```
Might be a better choice as it is more simpler.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,180 @@
...
+            ui.showLoadingError();
+            tasks = new TaskList();
+        }
+    }
+
+    /**
+     * Runs the duke program and parses user's input.
+     */
+    public void run() {
+        Scanner scn = new Scanner(System.in);
```
Tex-Tang: Should we change `scn` to `scanner` for a clearer name?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,225 @@
...
+        this.numTasks = 0;
+    }
+
+    /**
+     * This method adds a task to the given task-list.
+     *
+     * @param noType The input String with type removed.
+     * @param type The type of the task
+     */
+    public String addTask(String noType, String type, boolean done, boolean load)
```
Tex-Tang: Should we add comment for `done` and `load` also?

Tex-Tang (commented):
Overall it is very good. Code in `Event.java`, `Deadline.java` and `ToDo.java` are very clean. Just need to fix some minor issues.

src/main/java/duke/ToDo.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+/**
+ * A ToDo task.
+ */
+public class ToDo extends Task {
+
+    public ToDo(String title, boolean done) {
```
ReubenChay, line 8: boolean variable can start with isDone instead

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,46 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * A task with a specified deadline.
+ */
+public class Deadline extends Task {
```
ReubenChay, line 10: can't find much coding standard issues not already found by other reviewer. Really clear code with good use of OOP!

PR #300, opened 2022-08-23
[Darren Teo] iP

darrentde:
# ChadBot
I am ChadBot and my goal is to be the **biggest bot in town**

```Java
if (i.equals("ChadBot")) {
    System.out.println("Hello ChadBot 🤠");
}
```

Features
- Track task
- Save task
- Delete task

Future Plans
1. Become cognitive
2. Auto-pilot coding
3. ~~Enslave humanity~~

Improvements 
- [ ] Unknown commands causing exceptions
- [ ] More unit test
- [ ] CI/CD pipeline
- [ ] Better JavaDocs comments

Learn more about [chat bots](https://en.wikipedia.org/wiki/Chatbot)


src/main/java/chad/Parser.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param userInput user input
+     * @throws ChadException Thrown when invalid error occurs
+     * @throws IOException Thrown when helper file cannot be open
+     */
+    public static void readUserInput(ArrayList<Task> taskList, String userInput) throws ChadException, IOException {
+        String[] tempArr = userInput.split(" ");
+        String keyword = tempArr[0];
+
+        try {
+            switch (keyword) {
```
jovitaanderson: Good use of switch instead of if/else to prevent deep nesting 🙂

src/main/java/chad/task/Task.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Set task done attribute to false
+     */
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+
+    @Override
```
jovitaanderson, line 53: Maybe you could also include javadocs for  the toString() methods to let others know what is printed for the user?

src/main/java/chad/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+     * Add event task into task list
+     * @param tasks tasks arraylist of tasks
+     * @param userInput event task added by user
+     * @throws ChadException Thrown if description or date time is invalid
+     * @throws IOException Thrown when file cannot be opened in IOException
+     */
+    public static void addEventTask(ArrayList<Task> tasks, String userInput) throws ChadException, IOException {
+        String outputText = "Got it. I've added this task:\n";
+        String[] temp = userInput.replaceFirst("event", "").strip().split("/at");
+        String taskDescription = temp[0].strip();
```
jovitaanderson: Maybe you could consider combining line 101 and line 102 into one ? Since String[] temp is not used elsewhere 🤔

src/main/java/chad/Ui.java
```diff
@@ -0,0 +1,138 @@
...
+        outputText += " " + currentTask;
+        Storage.toggleMarkTaskInFile(taskID);
+        Utility.printText(outputText);
+    }
+
+    public static void searchTaskByKeyword(ArrayList<Task> tasks, String userInput) {
+        String keyword = userInput.replaceFirst("find", "").trim();
+        String baseText = "Here are the matching tasks in your list:\n";
+        String output = baseText;
+        int count = 1;
```
jovitaanderson: Perhaps you could consider changing the variable name to index instead of count, since the output would be printing the index of the task🤔

jovitaanderson (commented):
Looks good 👍🏼 Majority of your code follows the coding standard, good job 🙂 
LGTM

src/main/java/chad/Utility.java
```diff
@@ -0,0 +1,16 @@
+package chad;
```
JordanChua: Perhaps the method of printText can be abstracted to the UI class instead as a public static method to be used in TaskList as well

src/main/java/chad/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Add deadline task into task list
+     * @param tasks tasks arraylist of tasks
+     * @param userInput deadline task added by user
+     * @throws ChadException Thrown if description or date time is invalid
+     * @throws IOException Thrown when file cannot be opened in IOException
+     */
+    public static void addDeadlineTask(ArrayList<Task> tasks, String userInput) throws ChadException, IOException {
+        String outputText = "Got it. I've added this task:\n";
```
JordanChua: Maybe this outputText can be defined as a constant String instead since its used when adding an event as well, or abstracted to the UI class

src/main/java/chad/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+
+        if (dateTime.isEmpty()) {
+            throw new ChadException("The date of a deadline cannot be empty.");
+        }
+
+        Deadline newTask = new Deadline(taskDescription, dateTime);
+        tasks.add(newTask);
+
+        String strIsDone = newTask.getDone() ? "1" : "0";
+        Storage.writeToFile("D | " + strIsDone + " | " + newTask.getDescription()
```
JordanChua: Perhaps this can be abstracted to a method writeToFile() for both Events and Deadline since the code is reused

src/main/java/chad/Ui.java
```diff
@@ -0,0 +1,138 @@
...
+        String outputText = "";
+        outputText += "OK, I've marked this task as not done yet:\n";
+        Task currentTask = tasks.get(taskID);
+        currentTask.markAsUndone();
+        outputText += " " + currentTask;
+        Storage.toggleMarkTaskInFile(taskID);
+        Utility.printText(outputText);
+    }
+
+    public static void searchTaskByKeyword(ArrayList<Task> tasks, String userInput) {
```
JordanChua: I was wondering if there is a particular reason for implementing the method here instead of within the TaskList class

src/main/java/chad/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Add deadline task into task list
+     * @param tasks tasks arraylist of tasks
+     * @param userInput deadline task added by user
+     * @throws ChadException Thrown if description or date time is invalid
+     * @throws IOException Thrown when file cannot be opened in IOException
+     */
+    public static void addDeadlineTask(ArrayList<Task> tasks, String userInput) throws ChadException, IOException {
```
JordanChua: Maybe the IOException is not needed since it is will not be thrown in the body of the method

src/main/java/chad/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+        String outputText = "Got it. I've added this task:\n";
+
+        String[] temp = userInput.replaceFirst("deadline", "").strip().split("/by");
+        String taskDescription = temp[0].strip();
+        String dateTime = temp[1].strip();
```
JordanChua: Perhaps the scenario where the date is entered but in a wrong format can be considered as well since it would be meaningful to tell the user that the date format input is not the correct one

JordanChua (commented):
Overall the code is really easy to read, although there were some parts of code that were reused and could be abstracted away. Some meaningful errors and exceptions can also be considered :)

src/main/java/chad/Chad.java
```diff
@@ -0,0 +1,33 @@
...
+ * Main way for users to interact with Chadbot
+ */
+public class Chad {
+
+    /**
+     * Main Function of Chad
+     * @param args Reads input
+     * @throws IOException Thrown when helper file cannot be open
+     * @throws ChadException Thrown when invalid error occurs
+     */
```
alvintfl: Perhaps you can include an empty line between the description and parameter section?
Maybe you can also include punctuation behind each parameter description?
I noticed the same issues in several other places too.

src/main/java/chad/task/Todo.java
```diff
@@ -0,0 +1,14 @@
+package chad.task;
+
+/**
+ * chad.task.Todo task that is child of task
+ */
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+    }
```
alvintfl, line 10: Perhaps you can include a Javadoc comment for this constructor? So that users do not need to read the code to understand if the constructor does anything else.

src/main/java/chad/Ui.java
```diff
@@ -0,0 +1,138 @@
...
+        Utility.printText(outputText.toString().trim());
+    }
+
+    /**
+     * Marks the task at specific index as done
+     * @param tasks arraylist of tasks
+     * @param taskID index of task
+     * @throws IOException throws error if file can't be open in chad.Storage
+     */
+    public static void markTask(ArrayList<Task> tasks, int taskID) throws IOException {
```
alvintfl: Perhaps ```taskID``` can be written as ```taskId``` instead?

alvintfl (commented):
Overall, the coding standard is good except for some formatting issues with the Javadocs.
Keep up the good work! 👍

PR #307, opened 2022-08-23
[Sun Ruoxin] iP

vantemoon:
# Jarvis
> "Surprisingly I'm **NOT** related to that guy with a yellow stone on his head 🤡  "

Jarvis is here to help you remember every task you need to take note of.
- text-based and super easy to use ‼️ 
- different types of tasks supported

Getting started now 💥 
1. Download it from [here](https://github.com/vantemoon/ip)
2. Double-click it
3. And enjoy stress-free task management 🤩 

*Hello?* What are you waiting for?

Features:
- [x] Managing to-dos
- [x] Managing deadlines
- [x] Managing events
- [ ] Reminders (coming soon)

---
If you are also a Java programmer, feel free to use it to practice Java too. 
Here's the `main` method:
``` java
public static void main(String[] args) throws DukeException {
    }
```


src/main/java/duke/task/Todo.java
```diff
@@ -0,0 +1,47 @@
...
+        super(description, isDone);
+    }
+
+    /**
+     * Returns the type of the task.
+     *
+     * @return "T" representing to-do
+     */
+    @Override
+    public String getType() {
```
AugustDespair, line 34: I think this could be a static variable instead, since all Todo classes share the same Type, which is 'T'.

src/main/java/duke/command/FindCommand.java
```diff
@@ -0,0 +1,58 @@
...
+     * Gives feedback to the user.
+     *
+     * @param tasks the list of tasks
+     * @param ui the UI
+     * @param storage the storage
+     */
+    @Override
+    public String execute(TaskList tasks, Ui ui, Storage storage) {
+        ArrayList<Task> matches = tasks.find(keyword);
+//        ui.findMessage(matches);
```
AugustDespair: If I remember correctly, all comments should be indented based on the line of code below it, so the double slash should be right next to `ui.findMessage(matches);`

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,54 @@
+package duke.task;
+
+/**
+ * Encapsulation of an event.
+ *
+ * @author Sun Ruoxin
+ */
+public class Event extends Task {
+    /** The time of the event. */
+    protected String at;
```
AugustDespair, line 10: I think a variable name like `eventTime` would be more descriptive than `at`.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,73 @@
...
+import duke.DukeException;
+
+/**
+ * Encapsulation of a deadline.
+ *
+ * @author Sun Ruoxin
+ */
+public class Deadline extends Task {
+    /** The date of the deadline. */
+    protected LocalDate by;
```
AugustDespair, line 16: I think having the variable name be `deadline` or `deadlineDate` would be more descriptive than `by`.

AugustDespair (commented):
Your code is really organized and clean. Honestly, most of my comments are just nitpicks.

src/main/java/duke/task/Deadline.java
o-ohst, line 16: I agree that `deadline` would be a more appropriate name.

src/main/java/duke/task/Event.java
o-ohst, line 10: I agree that `eventTime` would be a more appropriate name.

src/main/java/duke/task/Todo.java
vantemoon: Hi, thanks for your suggestion! I do get your point on how all Todo instances return the same 'T' thus the method should be static. However, since the getType() methods in Todo, Deadline and Event are overwriting the abstract method in Task, they cannot be static as overriding depends on dynamic binding:(

PR #311, opened 2022-08-23
[Motwani Richa Vijay] iP

richavm14:
** Neo**
Neo frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/richavm14/ip/releases/tag/A-Jar).
2. download neo.jar
3. clock it to add your tasks

And it is **FREE!**

Features:

- [x] Managing tasks
- [x] Reminders (coming soon)


src/main/java/neo/MarkCommand.java
```diff
@@ -21,5 +21,9 @@ void complete(String tempi) throws NeoException, IOException {
         arrayLL.getTask(tempii-1).setIsDone(true);
         System.out.println("Nice! I've marked this task as done");
         System.out.println(arrayLL.getTask(tempii-1).toString());
+        stor.Writedata(arrayLL.getTask(0));
+        for(int i =1; i<arrayLL.arrayL.size(); i++) {
```
L1uY1jun: Convention to have a space around operators based on Java coding standard, this line of code should be written as:
`for (int i = 1; i < arrayLL.arrayL.size(); i++) {`

src/main/java/neo/UnMarkCommand.java
```diff
@@ -21,6 +21,10 @@ void complete(String tempi) throws NeoException, IOException {
         arrayLL.getTask(tempii-1).setIsDone(false);
         System.out.println("ohk I've marked this task as not done");
         System.out.println(arrayLL.getTask(tempii-1).toString());
+        stor.Writedata(arrayLL.getTask(0));
+        for(int i =1; i<arrayLL.arrayL.size(); i++) {
```
L1uY1jun: Similar issue for this line of code, should be written as: `for (int i = 1; i < arrayLL.arrayL.size(); i++) {` based on java coding standard

src/main/java/Parser.java
```diff
@@ -0,0 +1,71 @@
...
+                        throw new NeoException("sorry todo cannot be empty");
+                    }
+                    if (!arr[0].equals("todo") && !arr[0].equals("event") && !arr[0].equals("deadline")) {
+                        throw new NeoException("Sorry I don't know what that means");
+                    }
+                }
+            } catch (NeoException | IOException e) {
+                System.out.println(e.getMessage());
+            }
+        }
```
L1uY1jun: This section of text could be simplified much more with a 'switch' statement instead of 'if-else' statement
For example:
```java
if (arr.length < 0) {
    throw new NeoException("Sorry I don't know what that means");
} else {
    String command = arr[0];
    switch (command) {
    case "mark": 
        this.mark = new MarkCommand(ui, stor, arrayLL);
        mark.complete(arr[1]);
        break;
    case "unmark":
        this.unmark = new UnMarkCommand(ui, stor, arrayLL);
        unmark.complete(arr[1]);
        break;
    case "delete":
        this.delete = new DeleteCommand(ui, stor, arrayLL);
        delete.complete(arr[1]);
        break;
    case "deadline": 
        //add the other cases and code in
        //below until the default case
    default:
        //throw NeoException if the command string does not match any of the cases
        throw new NeoException("Sorry I don't know what that means");
}
```
read up on the switch statement documentation for more clarifications.

src/main/java/Ui.java
```diff
@@ -0,0 +1,9 @@
+public class Ui {
+    protected String inp;
+    public Ui() {}
+    public void printStart() {
+        System.out.println("");
+        System.out.println("Hello! I'm Neo");
+        System.out.println("What can I do for you?");
+    }
+}
```
L1uY1jun: The Ui (User interface) class is supposed to handle all the interactions with the user, suggestion for you to add all the `System.out.println(...)` code found in all the *Commands.java classes into this class as methods for Ui to handle everything.

L1uY1jun (commented):
There was a lot of instances where operators was found not being surrounded by a space character (Java coding standard), for example:
👎 practice: `a=(b+c)*d;`
👍 practice: `a = (b + c) * d;`
same goes for all other operators like `>, <, ||, &&` etc
Overall good effort put into the commits, keep it up!

src/main/java/neo/AddCommand.java
```diff
@@ -0,0 +1,49 @@
+package neo;
+import java.io.IOException;
+public class AddCommand extends Command {
+
+    private TaskList arrayLL;
```
Aloynz, line 9: Perhaps a more intuitive variable name here?

src/main/java/neo/AddCommand.java
```diff
@@ -0,0 +1,49 @@
...
+        if (type ==1) {
+            String arri[];
+            arri = tempi.split("/at ");
+            String temp2 = arri[0];
+            String temp3 = arri[1];
+
+            Event e = new Event(temp2, temp3);
+            System.out.println("Added: " + e.toString());
+            arrayLL.addTask(e);
+            stor.Storedata(e);
```
Aloynz: Should these line of code be placed after the if statements since it might look repetitive?

src/main/java/neo/Deadline.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        LocalDate d1 = LocalDate.parse(by);
+        return "[D]" + super.toString() + " (by: " + d1.format(DateTimeFormatter.ofPattern("MMM dd yyyy")) + ")";
```
Aloynz: I like how the toString() method is neatly implemented.

src/main/java/neo/Parser.java
```diff
@@ -0,0 +1,72 @@
...
+                        this.add = new AddCommand(ui, stor, arrayLL, type);
+                        add.complete(tempi);
+                    }
+                    if (arr.length == 1 && arr[0].equals("todo")) {
+                        throw new NeoException("sorry todo cannot be empty");
+                    }
+                    if (!arr[0].equals("todo") && !arr[0].equals("event") && !arr[0].equals("deadline")) {
+                        throw new NeoException("Sorry I don't know what that means");
+                    }
+                }
```
Aloynz: Should switch statement be used in placed of the multiple if statements?

Aloynz (commented):
Overall, I found the code easy to read except a few places where too many if-else statements were used.  You may wish to consider the use of switch statements in certain areas of the code.

src/main/java/Parser.java
richavm14: Understood, Will modify it 
Thank you

src/main/java/neo/AddCommand.java
```diff
@@ -0,0 +1,68 @@
...
+            System.out.println("Added: " + e.toString());
+            arrayLL.addTask(e);
+            stor.storeData(e);
+        }
+        if (type ==2) {
+            ToDo td = new ToDo(tempi);
+            System.out.println("Added: " + td.toString());
+            arrayLL.addTask(td);
+            stor.storeData(td);
+        }
```
zlimez: Hmm, since this is also parsing maybe could move this segment of the code to Parser?

src/main/java/neo/Neo.java
```diff
@@ -0,0 +1,53 @@
...
+        Neo neo = new Neo();
+        neo.ui.printStart();
+        Scanner sc = new Scanner(System.in);
+
+        while (true) {
+            System.out.println("Please enter items you want to add to the list, if you want to quit enter bye");
+            System.out.println("");
+            System.out.print("Enter Your command: ");
+            userText = sc.nextLine();
+            System.out.println("");
```
zlimez: Another thought, maybe this segment can be moved to the UI to make the responsibility of each class narrower?

zlimez (commented):
Just a few suggestions, and out of curiosity did you implement your gui using javascript jquery haha?

PR #316, opened 2022-08-23
[Aloysius Neo Ze En] iP

Aloynz:
### DukeLite

> If you embrace that the things that you can do are limitless, you can put your ding in the universe. 
You can change the world. - Tim Cook [(source)](https://github.com/Aloynz/ip/releases)

DukeLite is a **light weighted** application which helps you to keep track of tasks. It's,
- **simple** to use
- **rich** in features
- **_EASY_** to pick up

All you need to do is,
1.  download it from [here](https://github.com/Aloynz/ip/releases)
2. double click the .jar file to run
3. add your tasks.
4. let it track your tasks 😍

Features :

- [x]  Managing tasks
- [x] Managing deadlines, event
- [ ]  Display a GUI (coming soon)

What's a better way than not using DukeLite?

```ruby
public class Main {

         System.out.println("DUKELITE IS AWESOME!!!"); 
    }
}
```


src/main/java/seedu/duke/Duke.java
```diff
@@ -0,0 +1,74 @@
...
+public class Duke  {
+    
+    /*Storage object to handle reading and writing to hard disk */
+    private Storage storage;
+    /*Tasklist containing an arraylist of tasks */
+    private TaskList tasks;
+    /*Ui component to handle the user interface of the program */
+    private Ui ui;
+    /*Parser object to parse user input into recognised Commands */
+    private Parser parser;
```
charleslimjh:
```suggestion
    /** Storage object to handle reading and writing to hard disk */
    private Storage storage;
    /** Tasklist containing an arraylist of tasks */
    private TaskList tasks;
    /** Ui component to handle the user interface of the program */
    private Ui ui;
    /** Parser object to parse user input into recognised Commands */
    private Parser parser;
```
I think Javadocs for class members should still start with `/**` according to the coding standard. Might be good to take note?

src/main/java/seedu/duke/Command.java
```diff
@@ -0,0 +1,160 @@
...
+                if (input.indexOf('/') == -1) {
+                    throw new DukeException("The date of the event cannot be empty");
+                }
+
+                split = input.substring(6).split("/");
+                task = new Event(split[0], split[1].substring(3));
+                tasks.add(task);
+                ui.add(task);
+                output += "Now you have " + tasks.size() + (tasks.size() == 1 ? " task" : " tasks")
+                + " in the list.";
```
charleslimjh:
```suggestion
                output += "Now you have " + tasks.size() 
                        + (tasks.size() == 1 ? " task" : " tasks")
                        + " in the list.";
```

I like the line break to improve readability, but line indentation for wrapped lines should be 8 spaces. Maybe consider this change instead? Same change can be applied in other commands in `Command.java`.

src/main/java/seedu/duke/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public Deadline(String description, LocalDateTime date) {
+        super(description);
+        this.by = date;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + "(by: " + by.getMonth() + "-" + by.getDayOfMonth() + "-" + by.getYear()
+            + " " + String.format("%02d", by.getHour()) + ":" + String.format("%02d", by.getMinute()) + ")";
```
charleslimjh, line 52: Not sure I like how many characters are in each line here. While it does not hit the suggested 110-120 character limit, would it be better to break each line further ahead?

src/main/java/seedu/duke/Task.java
```diff
@@ -0,0 +1,60 @@
...
+        return "Nice! I've marked this task as done:\n[X] " + this.description;
+    }
+
+    /**
+     * Set the task to be uncompleted.
+     */
+    public String setUndone() {
+        this.isDone = false;
+        return "OK, I've marked this task as not done yet:\n[ ] " + this.description;
+    }
```
charleslimjh: Looks like a minor typo in Javadocs, the method description should start with a verb in plural form.
```suggestion
    /**
     * Sets the task to be completed.
     */
    public String setDone() {
        this.isDone = true;
        return "Nice! I've marked this task as done:\n[X] " + this.description;
    }

    /**
     * Sets the task to be uncompleted.
     */
    public String setUndone() {
        this.isDone = false;
        return "OK, I've marked this task as not done yet:\n[ ] " + this.description;
    }
```

charleslimjh (commented):
Overall, the code looks good and adheres to the coding conventions, save for minor typos here and there! 😄 

In certain areas, there were too many line breaks as well. It might be good to relook at some of your longer functions and remove any unnecessary whitespace to reduce the LoC in your project.

PR #318, opened 2022-08-23
[Sheryl Kong] iP

sherylkong18:
### Duke-it
 > Anything on your mind? Duke it!

Duke-it is ...
- text-based
- easy to learn

**HOW TO USE?**
1. Download it [HERE](https://github.com/sherylkong18/ip)
2. Simply _**Duke**_ your ToDo's off the top of your head anytime, anywhere.
3. View your list of tasks in one glance
- [ ] Check your tasks off as you complete them!

Best of all, it is **COMPLETELY FREE**! 😙

```java
public class Main {
    }
}

```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,48 @@
...
+                c.execute(taskList, storage);
+                isExit = c.isByeCommand();
+            } catch (DukeException de) {
+                System.out.println(de);
+            }
+        }
+
+    }
+
+    public static void main(String[] args) throws DukeException, IOException {
```
danielk0k: Should the main method be throwing exceptions if they are not caught?
Similar comment for line 18 as the DukeException that `storage.load()` throws has been caught.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Task class to represent a generic task.
+ *
+ * @author Sheryl Kong (A0240686Y)
+ */
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    private static int taskCount = 0;
```
danielk0k, line 12: Interesting way of keeping track of the number of tasks in the list :)
Would it be more appropriate to have this in the TaskList class? Or perhaps use the `arraylist.size()` method?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,161 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Scanner;
+
+import duke.command.*;
```
danielk0k, line 7: Kindly note that the coding standard does not allow "import all" statements.
> Imported classes should always be listed explicitly.

src/main/java/duke/UI.java
```diff
@@ -0,0 +1,70 @@
...
+* UI class to handle all user interactions and system output.
+*
+* @author Sheryl Kong (A0240686Y)
+*/
+
+public class UI {
+    public static final String DIVIDER = "––––––––––––––––––––––––––––––––––––––––––-––––––––––––––––––––––––-–––––––––––––––––––––––––––––––––";
+
+    public static void welcome() {
+        System.out.println("Hello! I am Sheryl. Please enter your command: ");
```
danielk0k: I like the personalised name you have given 👍🏻

danielk0k (commented):
Great work for week 3 🌟 ! Some minor edits and comments. But overall, the logic of the code is sound and I like that you made some changes to the default duke name.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,32 @@
...
+    private String date;
+
+    /**
+     * Constructor for Event class
+     *
+     * @param description description of the Event task
+     * @param isDone indicates if the task is marked as done
+     * @param date due date of the Event task
+     */
+
```
chao890, line 21: Perhaps you could remove the empty line between your javadocs and your method?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,74 @@
+package duke;
+
+import duke.command.Command;
+
+import java.io.*;
```
chao890, line 5: Perhaps we could only import the required classes?

src/main/java/duke/command/AddDeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public AddDeadlineCommand(String description, boolean isDone, String date) {
+        this.description = description;
+        this.isDone = isDone;
+        this.date = date;
+    }
+
+    @Override
+    public void execute(TaskList taskList, Storage storage) throws DukeException {
+        if (!storage.checkIsLoadingFile()) { this.date = Parser.parseDate(date); }
```
chao890: Perhaps you could use Egyptian style braces? 
```suggestion
        if (!storage.checkIsLoadingFile()) {
            this.date = Parser.parseDate(date);
        }
```

Same for line 34

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,161 @@
...
+
+            default:
+                throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
+    }
+
+    public static Command parseFileLine(String desc) throws DukeException {
+        String[] words = desc.split("\\s\\|\\s");
+        int size = words.length;
+        if (size == 1) return new NullCommand();
```
chao890, line 308: I noticed that your NullCommand() throws an exception that is not caught, and returns nothing.

Perhaps it will be more suitable to create a Duke exception that handles a command with no description, and remove the NullCommand class?

chao890 (commented):
LGTM! Other than some small coding standard violations, everything looks good. I also liked how you split your commands into different classes!

PR #322, opened 2022-08-23
[Benjamin Soh] iP

bensohh:
## DukePro

> "The Web as I envisaged it, we have not seen it yet. The future is still so much bigger than the past." - Tim Berners-Lee, Inventor of the World Wide Web [(source)](https://www.brainyquote.com/quotes/tim_bernerslee_409503)

DukePro frees your mind of having to remember things you need to do. It's

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://github.com/bensohh/ip).
2. double-click it.
3. add your tasks.

And it is **FREE!**

### Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)
##
```java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,64 @@
...
+ * An interactive ChatBot that provides functions of a to-do list
+ * depending on the user's input.
+ */
+public class Duke {
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Contructor to create an instance of Duke.
```
sharmaine1028: Spelling error: Constructor

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @param filePath the String representation of the relative file path to store the data
```
sharmaine1028: @param filePath String representation of the relative file path to store the data

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,64 @@
...
+        } catch (FileNotFoundException e) {
+            ui.showError(e.getMessage());
+            tasks = new TaskList();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    /**
+     * To kickstart the entire ChatBot.
```
sharmaine1028: Perhaps could write "Kickstarts the entire Chatbot" instead as first sentence should start in form of "Kickstarts...., Returns..., Sends..." etc.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public void run() {
+        System.out.println(ui.showWelcome());
```
sharmaine1028: Why not make ui.showWelcome() function directly print out the greeting? I feel that this will be more in line with the Tell, Don't Ask principle.

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,120 @@
...
+    private int taskType;
+    private String taskDetails;
+
+    /**
+     * Constructor to create an instance of AddCommand.
+     *
+     * @param taskType Integer value representing the type of task, 0-TODO Task,
+     *                 1-DEADLINE Task, 2-EVENT Task
+     * @param taskDetails String representation of the task description
+     * @return instance of AddCommand
```
sharmaine1028: I thought that constructor does not return anything in Java? No need for @return tag

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    public AddCommand(int taskType, String taskDetails) {
+        this.taskType = taskType;
+        this.taskDetails = taskDetails;
+    }
+
+    /**
+     * Create the corresponding task based on the taskType in AddCommand and adds task
```
sharmaine1028: Creates

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,120 @@
...
+     * to the TaskList. The task list is then stored on a file under data/tasks.txt
+     *
+     * @param tasks a list to add and keep track of the tasks
```
sharmaine1028: Should be capital letter after parameter, aka @param tasks List to add and keep track of the tasks

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,98 @@
...
+        //Exception: Throw an error when user tries to delete from an empty list
+        if (tasks.isEmpty()) {
+            String errorMessage = "__________________________________________________\n"
+                    + "OOPS!!! There are no task left to be deleted!\n"
+                    + "__________________________________________________";
+            throw new DukeException(errorMessage);
+        }
+
+        //Exception: Throw an error when the second half after "delete" keyword is greater than task_list
+        if (taskNumber > tasks.size() || taskNumber < 1) {
```
sharmaine1028, line 49: Why is there taskNumber < 1 condition in this line when it is already handled above?

sharmaine1028: I'm assuming that taskNumber < 1 is the same check as isEmpty() check

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return false since a DeleteCommand does not end the ChatBot
+     */
+    public boolean isExit() {
+        return false;
+    }
+
+    /**
+     * A String representation of successfully executing the DeleteCommand.
+     *
+     * @return a String to notify the user that a task has been removed
```
sharmaine1028: Remember to capitalise after @return tag as well

src/main/java/duke/command/UnmarkCommand.java
```diff
@@ -0,0 +1,90 @@
...
+    private int taskNumber;
+
+    /**
+     * Constructor to create an instance of UnmarkCommand.
+     *
+     * @param taskDetails a String representation of the task descriptions
+     * @return an instance of UnmarkCommand
+     */
+    public UnmarkCommand(String taskDetails) {
+        this.taskNumber = Integer.parseInt(taskDetails);
```
sharmaine1028: I'm not very sure if possible error that can be thrown by parseInt has been handled. For instance if user types in unmark hello when it should be unmark 1. I checked the Parser class and it does not seem to have handled this case. Perhaps can consider throwing some exception here or somewhere higher in the chain.

sharmaine1028 (commented):
In general, looks pretty good. Import statements are in the right order for the files I saw, static imports first etc. Indentation and Egyptian style braces are also in place.

However, take note of the conventions for JavaDocs
1. First sentence as short summary of method, should be in present tense form "Adds..., Returns..., Sends..."
2. First letter for parameter description should be capitalised, E.g. @param zone Zone of position

Other than that, I think the Java coding standards were largely followed. Good job!

src/main/java/duke/command/DeleteCommand.java
bensohh: Hi! Sorry for the confusion, the isEmpty() check is to check if the tasks (TaskList) is empty while taskNumber here refers to the input from the user, in the event when the tasks (TaskList) contains task and they enter delete 0, delete -1 or any number that is < 1 we will throw an error since there can only be taskNumber starting from 1 onwards. Thank you for taking your time to review.

src/main/java/duke/command/UnmarkCommand.java
bensohh: Thank you for catching something I missed out! I will be adding in a NumberFormat exception when parsing the inputs from the user to ensure that the command is correct.

PR #323, opened 2022-08-23
[Gujar Parth Shailesh] iP

parth-io:
## Duke has more tasks now!

> Welcome to the new age

It now supports these features and more:

- **adding** todos, events or deadlines; ~~deleting~~ tasks; _searching_ for tasks
- ___easier___ to use, with better error messages
- persists data in local storage

Steps to use :bowtie::
1. Download it from [here](https://github.com/parth-io/ip/releases/download/A-Jar/duke.jar)
2. Double-click it or run `$ java -jar duke.jar` in your bash terminal from the folder where the file is.
3. Add your tasks!

Things to improve:
- [x] Code conventions
- [ ] Better use of OOP principles

```java
public void printGreeting() {
        String logo = " ____        _        \n"
                + "|  _ \\ _   _| | _____ \n"
                + "| | | | | | | |/ / _ \\\n"
                + "| |_| | |_| |   <  __/\n"
        System.out.println("Hello from\n" + logo);
    }
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public void run() {
+        ui.showWelcome();
+        Parser parser = new Parser(taskList);
+        while (true) {
+            try {
+                String nextCommand = ui.getNextCommand();
+                ui.showLine();
+                System.out.print(parser.parseUserCommand(nextCommand, parser.breakLoopIndicator));
+                if (parser.breakLoopIndicator.getIsExitCommand()) {
```
riccqi:
```suggestion
                if (parser.breakLoopIndicator.getExitCommand()) {
```
small nit, i think we can drop the `is`

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,60 @@
...
+                    ui.showExitMessage();
+                    break;
+                }
+//                Command c = Parser.parse(fullCommand);
+//                c.execute(tasks, ui, storage);
+            } catch (CustomMessageException e) {
+                ui.showError(e.getMessage());
```
riccqi: could we have taken advantage of the `toString` method in your exception class instead?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,167 @@
...
+        Command taskType;
+        try {
+            taskType = Command.valueOf(commands[0].toUpperCase());
+        } catch (IllegalArgumentException e) {
+            throw new CustomMessageException((" ☹ OOPS!!! I'm sorry, but I " +
+                    "don't know what that means :-(\n"));
+        }
+        commands[0] = "";
+        int index;
+        String toPrint;
```
riccqi: i think would be better to name it a noun, since its a variable

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,59 @@
...
+
+    //this is only public for now, need to refactor Parser into Command classes and call this from Command instead
+    /**
+     * Returns an indented message.
+     * @param message The message to indent
+     * @return A {@code String} representing the indented message
+     */
+    public static String indentedMessage(String message) {
+        return "     " + message;
+    }
```
riccqi: this is a function, so i would do `indentMessage` instead!

src/test/java/duke/TaskListTest.java
```diff
@@ -0,0 +1,65 @@
...
+    static void tearDown() {
+        if (file.exists() && !file.delete()) {
+            throw new RuntimeException("Could not delete file");
+        }
+    }
+
+    /**
+     *
+     */
+    @BeforeEach
```
riccqi, line 20: Good use of the junit annotations!

riccqi (commented):
LGTM, some small nits to fix!

src/main/java/duke/Duke.java
parth-io: Alright, thanks for the comments

PR #331, opened 2022-08-23
[Charles Lim Jun Hao] iP

charleslimjh:
# Duke

> DukeBot is a **CLI** application that helps *manage your tasks*. 👍 

You can:
- Add tasks
- Delete tasks
- Mark tasks as complete/incomplete

To get started:
1. Download the `.jar` file from [here](https://github.com/charleslimjh/ip/releases/tag/A-Jar).
2. Navigate to the file in your terminal.
3. Run the following command `java -jar duke.jar`
4. Profit!

Features to implement:
- [x] Basic CRUD
- [x] Load to file
- [ ] GUI
- [ ] Testing 

Here's a preview of the main code:
```java
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,247 @@
...
+                this.addEvent(userInput);
+            } else if (userInput[0].equals("mark")) {
+                this.mark(userInput);
+            } else if (userInput[0].equals("unmark")) {
+                this.unmark(userInput);
+            } else if (userInput[0].equals("delete")) {
+                this.delete(userInput);
+            } else {
+                throw new DukeException("Unrecognized command.");
+            }
```
sohenze: Perhaps a switch statement would be better here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,247 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Scanner;
+import java.util.stream.Stream;
+
 public class Duke {
+    private ArrayList<Task> db = new ArrayList<>(10);
```
sohenze: Perhaps a more intuitive variable name here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,247 @@
...
+            switch (tmp[0]) {
+                case "T":
+                    db.add(new Todo(tmp));
+                    break;
+                case "D":
+                    db.add(new Deadline(tmp));
+                    break;
+                case "E":
+                    db.add(new Event(tmp));
+                    break;
```
sohenze: I like how you used a switch statement here! But take note that we are not supposed to have indentations for case clauses!

src/main/java/Duke.java
charleslimjh: Yup, just implemented it. Good suggestion!

src/main/java/Duke.java
charleslimjh: Personally, I thought `db` was a very intuitive name since it commonly refers to 'database'. A good point to consider though, thank you!

PR #333, opened 2022-08-23
[Ong Han Wei] iP

rexong:
### DukeBot

> “Your mind is for having ideas, not holding them.” – David Allen [(source)](https://www.google.com/)

DukeBot is your personal task bot that helps you remember your task! It's
- Text-based
- CLI
- ~~easy~~ Very Easy to use

All you need to do is 

1. download from the link [here](https://github.com/rexong/ip/releases/tag/A-Jar)
2. Double click duke.jar
3. Run the file
4. Add your task

It's **Free**! :smiley:

Features:

- [x] Managing tasks
- [ ] Managing Deadline (Coming soon) 
- [ ] Reminders (Coming soon)

Here is a random code in Java,
``` java
public class Main {
        System.out.println("Hello world!");
    }
}
```
_Here is a random inline code._
`
System.out.println("Duke");
`


src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,207 @@
...
+            return prepareDelete(commandWord, description);
+        case FindCommand.COMMAND_WORD:
+            return prepareFind(commandWord, description);
+        default:
+            return prepareInvalid();
+        }
+    }
+
+    private static String getCommandWord(String command) {
+        if (command.startsWith("list")) {
```
Jonaspng, line 66: Could consider switching to switch and case statements?

src/main/java/duke/util/TaskList.java
```diff
@@ -0,0 +1,203 @@
...
+        if (text != "") {
+            String[] texts = text.split("\n");
+            String taskType;
+            String description;
+            for (int i = 0; i < texts.length; i++) {
+                taskType = texts[i].substring(3,6);
+                String[] descriptions;
+                switch (taskType) {
+                case "[T]":
+                    description = texts[i].substring(10);
```
Jonaspng: Maybe you could consider avoiding the magic number 10. You could consider using a named constant instead.
image

src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Represents the User Interface of duke.
+ * Provides a welcome and goodbye text to be display in the command line.
+ */
+public class Ui {
+    private final String DUKE_INTRODUCTION = "Hello, I'm your personal assistant, Duke.";
+    private final String DUKE_HELP = "How can I assist you today?";
+    private final String DUKE_END = "Pleasure to be at your service! Run me again if you need more assistance! :)";
+    private final String divider = "__________________________________________________________";
```
Jonaspng: Shouldn't this be named like a constant?

Jonaspng (commented):
Looks good to me. Just some minor issues to fix.

src/main/java/duke/util/Parser.java
rexong: was considering switch statements initially, however, as I'm using the method `startsWith`, I couldn't implement switch statement as the case for switch statement requires a string but the method `startsWith` return a boolean. If I want to implement switch statement, then I probably have to filter out the command word from the argument `command`.

src/main/java/duke/util/TaskList.java
rexong: Ah right, thanks for pointing this out!

src/main/java/duke/util/Ui.java
rexong: yup, I think I missed this. Thanks!

src/main/java/duke/util/Parser.java
Jonaspng: Wouldn't it be worth the trouble?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,92 @@
...
+     * Marks task as done.
+     *
+     * @throws TaskMarkException when task is already done.
+     */
+    public void mark() throws TaskMarkException{
+        if (isDone) {
+            throw new TaskMarkException();
+        }
+        isDone = true;
+    }
```
Jiacheng-y: Since this is a setter method for a Boolean variable, perhaps you can consider naming it `markDone()`

src/main/java/duke/util/DateTimeHandler.java
```diff
@@ -0,0 +1,45 @@
...
+     * Formats and stores dateTime as a LocalDateTime object.
+     *
+     * @param strDateTime Date and Time in the format dd/MM/yyyy HHmm.
+     * @return a LocalDateTime object.
+     * @throws UnexpectedDateTimeFormatException when strDateTime is not in dd/MM/yyyy HHmm format.
+     */
+    public static LocalDateTime formatDukeDateTime(String strDateTime) throws UnexpectedDateTimeFormatException {
+        try {
+            LocalDateTime dateTime = LocalDateTime.parse(strDateTime, dukeDateTimeFormat);
+            return dateTime;
```
Jiacheng-y: I really like that you've split up the DateTimeFormatter variables out of the parse function! The code looks really neat! :)

src/main/java/duke/util/Parser.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private static String getDescription(String command) {
+        int indexOfFirstWhiteSpace = command.indexOf(' ');
+        return (indexOfFirstWhiteSpace != -1)
+                ? command.substring(indexOfFirstWhiteSpace + 1)
+                : "";
+    }
+
+    private static int convertStringToInt(String number) throws IntegerExpectedException{
```
Jiacheng-y: perhaps there should be a white space before the open braces : 

```java
private static int convertStringToInt(String number) throws IntegerExpectedException {
```

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param fileName Name of the file to read and write to.
+     */
+    public Storage(String fileName) {
+        dukeFile = new File(fileName);
+    }
+
+    /**
+     * Shows whether file is present or not.
+     * @return True if file is present else false.
```
Jiacheng-y, line 33: perhaps `true` should be in lowercase since that's how the boolean `true` is represented in java

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    /**
+     * Updates the file with the new contents in TaskList.
+     *
+     * @param ui Ui object to display message from storage
+     * @param obj Object to be written into the file.
+     */
+    public void update(Object obj, Ui ui) {
```
Jiacheng-y: it may not be a wise choice for a public method to have a parameter with the most general class Object. I notice that it doesn't break your code or anything since every object has a `toString()` method but this allows anyone importing the Storage class to write anything into the file. I'm not sure that's a good idea.

Jiacheng-y (commented):
Generally looks good to merge! There are just a few nits to fix. The code is really well segregated and the classes and functions are quite comprehensive. I've learnt a lot from reading your code! :)

PR #334, opened 2022-08-23
[Jason Christopher] iP

jasonchristopher21:
# Shiba 🦮 

> "So task, much wow" - _Shiba Doge_

Shiba is a task management software dedicated to help you organise your todos, deadlines and events! It's
- Straightforward
- Minimalistic
- ~Easy~ _VERY EASY_ to use

To use **Shiba**, you need to:
1. Download `Duke.jar` file from [here](https://github.com/jasonchristopher21/nus-cs2103t-ip/releases/tag/A-Jar)
2. Run `java -jar a Duke.jar` in your terminal
3. Add your tasks!

Features: 
- [x] Add todo, deadline and event
- [ ] Mark your tasks as done
- [ ] Remove your tasks when needed

If you're a Java Programmer, you may want to modify the Shiba code! Here's the main method:
```java
    Duke duke = new Duke();
    duke.run();
}
```


src/main/java/Date.java
```diff
@@ -0,0 +1,67 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+public class Date {
+
+    protected LocalDate date = LocalDate.now();
+    protected int time;
+
+    public Date(String str) throws DukeInvalidDateException {
+        String[] splitted = str.split("\\s+");
```
LJXSean: Perhaps you could use a plural form to distinguish a multi-valued variable?

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+                        Integer index;
+                        if (splitted.length == 1) {
+                            throw new DukeUnspecifiedNumberException("mark");
+                        }
+                        try {
+                            index = Integer.valueOf(splitted[1]);
+                            tasks.markAsDone(index);
+                        } catch (NumberFormatException err) {
+                            throw new DukeInvalidNumberFormatException("mark");
+                        }
```
LJXSean: Is it possible to avoid deep nesting with more than 3 levels of indentation?

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+                }
+            } catch (DukeException ex) {
+                drawLine();
+                indentMessage(ex.getMessage());
+                drawLine();
+            }
+        }
+    }
+
+    private static void greetingMessage() {
```
LJXSean: Would using a verb for the method name would be more suitable?

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+        drawLine();
+    }
+
+    private static void echo(String msg) {
+        drawLine();
+        indentMessage(msg);
+        drawLine();
+    }
+
+    public static void indentMessage(String msg) {
```
LJXSean: Using a verb here might be better as well!

src/main/java/Storage.java
```diff
@@ -0,0 +1,154 @@
...
+                if (splitted[1] != null) {
+                    if (splitted[1].startsWith("todo")) {
+                        t = addTodo(splitted[1]);
+                    } else if (splitted[1].startsWith("deadline")) {
+                        t = addDeadline(splitted[1]);
+                    } else if (splitted[1].startsWith("event")) {
+                        t = addEvent(splitted[1]);
+                    } else {
+                        t = new Task("dummy");
+                    }
```
LJXSean: Could you avoid using arrowhead nesting if possible?

LJXSean (commented):
LGTM, with some nits to fix! 👍

src/main/java/Storage.java
```diff
@@ -0,0 +1,154 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+class Storage {
+
+    private File f;
+    private static final String path = "src/main/data/userTasks.txt";
```
johnrhimawan: Should this be in uppercase instead since you are declaring a constant?

```suggestion
    private static final String PATH = "src/main/data/userTasks.txt";
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+    }
+
+    private static void greetingMessage() {
+        drawLine();
+        indentMessage("Hello! I'm Duke");
+        indentMessage("What can I do for you?");
+        drawLine();
+    }
+
+    private static void bye() {
```
johnrhimawan: Perhaps the method name could use a verb like ```sayGoodbye()```? I noticed the same issue in other places as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
...
+        isClosed = true;
+        drawLine();
+        indentMessage("Bye, Hope to see you again soon!");
+        drawLine();
+    }
+
+    private static void echo(String msg) {
```
johnrhimawan: Perhaps a more intuitive method name could be used here?

src/main/java/Storage.java
```diff
@@ -0,0 +1,154 @@
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.BufferedWriter;
```
johnrhimawan: Should ```java.io.BufferedWriter``` be moved to before ```import java.io.File``` to ensure consistent ordering?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,21 @@
+public class Deadline extends Task {
```
johnrhimawan: Would it be beneficial to organize your code into packages? I noticed the same issue in other files as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,108 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    private static final String LINE = "    __________________________________________________________________________";
+    private static final String INDENT = "     ";
+
+    private static boolean isClosed = false;
+    private static TaskList tasks = new TaskList();
+    private static int counter = 0;
```
johnrhimawan: Perhaps this variable could be removed since it is not used anywhere else?

johnrhimawan (commented):
Overall, your code was easy to read for the most part. I noticed that you followed the coding standard closely in your work. However, there were also several instances where you did not use a verb to name a method and you did not organize your code into packages. Otherwise, LGTM and just a few nits to fix. Excellent work!

src/main/java/Storage.java
jasonchristopher21: Sure thing, thank you very much for your comments, really appreciate them. Will certainly take your comments into consideration for the next iterations! :smile:

src/main/java/Storage.java
jasonchristopher21: Right, I guess I missed this out. Thank you for your comment!

src/main/java/Duke.java
jasonchristopher21: Right, thank you for your comment, will implement the change in the next iteration.

src/main/java/Duke.java
jasonchristopher21: Right, thank you for your comment, will implement the change in the next iteration.

src/main/java/Storage.java
jasonchristopher21: Right, I guess I missed this out when sorting it alphabetically. Thank you for your comment!

src/main/java/Deadline.java
jasonchristopher21: I agree with that. I have implemented them under the `A-Packages` tag in the current iteration (as of 9th September), since I had not implemented it last week. Thank you for your comment!

src/main/java/Duke.java
jasonchristopher21: Cool, I'll consider removing it for the next iterations.  
Thank you for your comments, I really appreciate them. I'll take them into consideration and implement the changes in the next iteration :smile:

src/main/java/Duke.java
jasonchristopher21: Yup, I just caught up on the SLAP method and the arrowhead style, I would consider implementing it in the next few iterations. Thank you for your comment.

src/main/java/Duke.java
jasonchristopher21: Right, I'll take it into consideration. Thank you for your comment.

src/main/java/Duke.java
jasonchristopher21: Right, I'll take it into consideration. Thank you for your comment.

src/main/java/Date.java
jasonchristopher21: Right, I'll probably consider renaming it to `splittedWords` instead. Thank you for your comment.

PR #336, opened 2022-08-23
[Livia Michelle Iskandar Leo] iP

liviamil:
# Sally, your personal task manager bot

> "This will be done no matter what" - Livia ([img](https://images.app.goo.gl/XALxLgjZCMhucucV8))

Sally frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ **__SUPER FAST__** to use

All you need to do is,
1. download it from [here](github.com/liviamil/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [X] Managing tasks
- [X] Managing deadlines
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. here's the `main` method:
```ruby
public class Main {
    }
}
```

Some of my most used Git commands so far:
```
git status
git add
git commit
```


data/Sally.txt
```diff
@@ -0,0 +1,5 @@
+T | 1 | read book
```
neosunhan: Perhaps you can consider adding data files to `.gitignore`. Otherwise users who clone your repo may end up with tasks loaded that they did not add.

src/main/java/Event.java
```diff
@@ -0,0 +1,13 @@
+public class Event extends Task {
+    String moreInfo;
```
neosunhan: Missing access modifier.

src/main/java/Sally.java
```diff
@@ -0,0 +1,291 @@
...
+        try {
+            readsFile();
+        } catch (FileNotFoundException e) {
+        }
+        sc = new Scanner(System.in);
+        messaging();
+    }
+
+    public static void readsFile() throws FileNotFoundException {
+        String filePath = "D:/NUS/Y2/S1/CS2103T/ip/data/Sally.txt";
```
neosunhan: Consider making `filePath` system-agnostic. Also, other users who run your app may not have a D drive.

src/main/java/Sally.java
```diff
@@ -0,0 +1,291 @@
...
+                throw new SallyException.SallyInvalidInputException();
+            }
+        } catch (SallyException e) {
+            System.out.println(e);
+        }
+        return Task.Type.TODO;
+    }
+
+    public static void savesFile() throws IOException {
+        String filePath = "D:/NUS/Y2/S1/CS2103T/ip/data/Sally.txt";
```
neosunhan: Consider storing `filePath` as a class field. This will make changing `filePath` easier as you only need to change it in one place.

src/main/java/Sally.java
```diff
@@ -0,0 +1,291 @@
...
+                }
+            }
+        } catch (SallyException e) {
+            System.out.println(e);
+        } finally {
+            messaging();
+        }
+    }
+
+    protected static String printList() {
```
neosunhan: Since this function does not actually do any printing, maybe you can rename it to `convertListToString` or something similar?

neosunhan (commented):
I've left some comments about the code. Regarding the storage of tasks, perhaps you can consider how other users with different machines (with different OS and file systems) will interact with your app.

src/main/java/sally/parser/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+
+import sally.command.AddCommand;
+import sally.command.ByeCommand;
+import sally.command.Command;
+import sally.command.DeleteCommand;
+import sally.command.MarkCommand;
+import sally.command.UnmarkCommand;
+import sally.exception.SallyException;
+import sally.storage.Storage;
+import sally.task.*;
```
ytingtan: Imported packages should be listed explicitly, maybe you could write down the attributes and methods you need from the `task` package and import in one-by-one? :) Otherwise LGTM!

src/main/java/sally/parser/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+                }
+            }
+
+            // Any other messages
+            else {
+                throw new SallyException.SallyInvalidInputException();
+            }
+        }
+
+        return new ByeCommand(); // Perlu diganti
```
ytingtan: Oops! Gentle reminder that all comments should be written in English ^_^

src/main/java/sally/task/Task.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+
+    public boolean getDoneStatus() {
```
ytingtan, line 71: As much as possible, it would be nice to name booleans to sound like booleans. Perhaps you could name this `isDone` instead? :)

ytingtan (commented):
Overall, I enjoyed reading your code as it was structured nicely and was aligned to the coding standard. I've made comments about some minor points, but overall LGTM! 👍 😄

PR #337, opened 2022-08-24
[Rachel Angelyn Gunawan] iP

Rachel-AG:
# Duke But Better

### Duke offers simple task management that is:
- text-based
- easy to use
- ~fast~ very FAST

### All you need to do is:
1. Download the latest version from [here](https://github.com/Rachel-AG/ip/releases)
2. Double click the jar file
3. Add your tasks, deadlines, and events!
4. Let Duke remember it for you 😄 

The best thing is... **IT'S FREE**

### Features:
- [x] Manage your tasks
- [ ] Reminders (coming soon)

```
        new Duke("data/duke.txt").run();
    }
````

### CODE REVIEW

File: src/main/java/duke/command/ByeCommand.java [hunk 1]
```diff
@@ -0,0 +1,16 @@
+package duke.command;
+
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+
```
lizekai-richard, line 5: I think adding a header here would be better 🤔

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Initializes an instance of a Command.
+     *
+     * @param isTerminator Indicates whether the command is a terminator.
+     */
+    protected Command(Boolean isTerminator) {
+        this.isTerminator = isTerminator;
+    }
+
+    public boolean isTerminatorGetter() {
```
lizekai-richard: Perhaps your method name should be isTerminator?

lizekai-richard (commented):
I like the way you follow coding standard overall 👍

src/main/java/duke/command/Command.java
Rachel-AG: Yeah, I think I should change my method name using verbs. Maybe something like isTerminator or getIsTerminator.

src/main/java/duke/command/ByeCommand.java
Rachel-AG: Yep, will add more headers later. Thank you for the feedback! :)

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,36 @@
+package duke.command;
+
+import duke.DukeException;
+import duke.storage.Storage;
+import duke.task.TaskList;
+
+/**
+ * The Command class represents the command entered by user.
+ */
+public abstract class Command {
```
Ugholaf, line 10: I like your use of abstract class here to abstract the different types of commands!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,116 @@
...
+            String[] eventArgs = getEventArgs(commandArg);
+            return new EventCommand(new Event(eventArgs[0], parseDateArg(eventArgs[1])));
+        case "bye":
+            if (!commandArg.equals("")) {
+                throw new DukeException("bye must not have any argument.");
+            }
+            return new ByeCommand();
+        default:
+            throw new DukeException("Sorry, I don't understand what that means :(");
+        }
```
Ugholaf: I like your use of `switch case` statement which makes your code more readable.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param taskDescription Specified task description.
+     */
+    protected Task(String taskDescription) {
+        this.taskDescription = taskDescription;
+        this.isDone = false;
+    }
+
+    public void isDoneSetter(boolean isDone) {
+        this.isDone = isDone;
+    }
```
Ugholaf: Missing Javadoc comment here.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * Returns the n-th task in the TaskList.
+     *
+     * @param n Index of task to be retrieved.
+     * @return n-th task in TaskList.
+     * @throws DukeException If n - 1 is outside the bounds of TaskList's size.
+     */
+    public Task getTaskN(int n) throws DukeException {
```
Ugholaf: Would it be better to name this method `getTaskByIndex` as N is confusing to readers what it refers to?

Ugholaf (commented):
Overall I like how you have structured your code and your naming conventions. The JavaDocs are also written very well! Good coding convention and code quality which makes the whole code very easy to read and understand!

src/main/java/duke/task/TaskList.java
Rachel-AG: That's true! I will rename the method in future commits

PR #340, opened 2022-08-24
[Wu Changjun] iP

Ugholaf:
# Duke
> "Your mind is for having ideas, not holding them." - David Allen [(source)](https://dansilvestre.com/productivity-quotes/)

- text-based
- easy to learn
- ~~FAST~~  SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com/Ugholaf/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Managing deadlines (coming soon!)
- [ ] Reminders (coming soon!)

---

``` Java
public class Duke {
        new Duke().run();
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,230 @@
...
+            default: {
+                throw new DukeException("OOPS!!! Invalid task type.\n");
+            }
+        }
+
+        taskList.add(task);
+
+        storage.appendTaskToFile(task);
+
+        String addTaskMessage =  "Got it. I've added this task:\n" +
```
DanielLimWeiEn: Would it be a good idea to extract out the printing of the messages to a separate class, perhaps a Ui class? since the printing of messages would fit in a Ui related class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,230 @@
...
+
+            String str = "Ok, I've marked this task as not done yet:\n" +
+                    task + "\n";
+            printTextWithDivider(str);
+        } catch (NumberFormatException | IndexOutOfBoundsException e) {
+            throw new DukeException("OOPS!!! The task index specified is not valid.\n");
+        }
+    }
+
+    private static void deleteTaskCommand(String[] inputs) throws DukeException {
```
DanielLimWeiEn: I like your naming of methods 👍. The names are very descriptive and very clearly convey what the methods are doing.

src/main/java/Duke.java
```diff
@@ -1,10 +1,230 @@
...
+            throw new DukeException("OOPS!!! The task index cannot be empty.\n");
+        }
+        try {
+            // Tasks are stored as 0-index but display as 1-index
+            // Minus 1 to get the correct task in the taskList
+            int taskIndex = Integer.parseInt(inputs[1]) - 1;
+            Task task = taskList.get(taskIndex);
+            task.markAsDone();
+            storage.writeAllTasksToFile(taskList);
+
```
DanielLimWeiEn: I'm not sure I like how the code here aims to do multiple things all at once. From carrying out the task to printing to the user interface. Do you think it is a good idea to maybe extract some of similar logic to separate classes?

src/main/java/Duke.java
```diff
@@ -1,10 +1,230 @@
...
+     *
+     * @param command The specified command
+     */
+    private static void executeCommand(String command) {
+        // Limit the words to 2
+        String[] inputs = command.split(" ", 2);
+
+        try {
+            switch (inputs[0]) {
+                // List out all tasks
```
DanielLimWeiEn: I think that the switch case here isn't following the coding standard they've given to us namely the standard below with the indentation.

```java
switch {
case 1:
    // ...
case 2:
    // ...
default:
    break;
}
```

DanielLimWeiEn (commented):
Overall looks pretty good 👍. I like your naming of methods and variables as it is very obvious what the methods and variables are for. There were however some areas that could be done better to make the code better, such as extracting out some functionality.

PR #346, opened 2022-08-24
[Lin Leyi] iP 

lleyi0606:
## Dukie

> “It takes as much energy to wish as it does to plan.” – Eleanor Roosevelt ([Source](https://www.brainyquote.com/quotes/eleanor_roosevelt_379411))

Your one-stop solution to record and manage your daily tasks. It's

- text-based
- efficient 
- straightforward to use! 

You just need to, 

1. download it from [here](https://github.com/lleyi0606/ip)
2. open it with your favourite code editor
3. input your tasks
4. sit back and let Dukie ~~do~~ remember them 🐥
 
### Features
- [ ] managing different types of tasks (todo, event, deadline) 
- [ ] sorting by date
- [ ] reminders _(coming soon)_ 

If you are a Java Programmer too, you can use it to practice programming! 
```ruby
        new Duke("data/duke.txt").run();
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,103 @@
...
+                case "todo":
+                case "deadline":
+                case "event":
+                    try {
+                        this.tasks.addTask(userInput);
+                    } catch (DukeException e) {
+                        ui.showMessage(e.getMessage());
+                    }
+                    break;
+                default:
```
feliciagan: Case clauses should not be indented

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,53 @@
...
+        return userInput.substring(9);
+    }
+
+    public static String parseUserDate(String userInput) {
+        if (userInput.contains("/")) {
+            System.out.println("Date: " + userInput.substring(userInput.indexOf("/") + 4));
+            return userInput.substring(userInput.indexOf("/") + 4);
+        }
+        return "";
+    }
```
feliciagan: If possible, could include javadoc comment for this method (and other methods in the other files)

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,142 @@
...
+*/
+enum TaskType {
+    TODO, DEADLINE, EVENT;
+}
+
+/*
+Task object that gets created by the user
+*/
+public class Task {
+    private boolean done;
```
feliciagan: Perhaps a more boolean-sounding name like isDone?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,142 @@
...
+            this.date = LocalDate.parse(timing);
+        }
+    }
+
+
+    public String getTask() {
+        return this.taskName;
+    }
+
+    public String isDoneString() {
```
feliciagan, line 65: Maybe a method name like getIsDoneString()? (since names representing methods must be verbs)

feliciagan (commented):
looks good!

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,142 @@
...
+            this.taskType = TaskType.TODO;
+            this.taskName = taskName;
+            this.date = null;
+            this.time = null;
+        } else if (taskType.equals("DEADLINE") || taskType.equals("EVENT")) {
+            if (taskType.equals("EVENT")) {
+                this.taskType = TaskType.EVENT;
+            } else {
+                this.taskType = TaskType.DEADLINE;
+            }
```
phualiting: Would it be better to remove the inner if else block, so that it is not checking for deadline and event twice, and will be clearer?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,142 @@
...
+Possible types of tasks to be created
+*/
+enum TaskType {
+    TODO, DEADLINE, EVENT;
+}
+
+/*
+Task object that gets created by the user
+*/
+public class Task {
```
phualiting, line 18: I think it would be clearer if you add additional Todo, Deadline and Event classes that extend Task, as I feel the checking of tasktype can be quite confusing?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,103 @@
...
+                    ui.showSchedule(date, list);
+                    break;
+                case "find":
+                    String keyword = Parser.getTaskName(userInput);
+                    ArrayList<Task> foundTasks = this.tasks.find(keyword);
+                    //ui.showMessage("Here are the matching tasks in your list:\n");
+                    //ui.showList(foundTasks);
+                    ui.showFoundTasks(foundTasks, keyword);
+                case "todo":
+                case "deadline":
```
phualiting: Are these missing the implementation?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,103 @@
...
+
+        Scanner myObj = new Scanner(System.in);  // Create a Scanner object
+        String userInput = "0";
+
+        while (true) {
+
+            userInput = myObj.nextLine();  // Read user input
+            String command = Parser.parseUserInput(userInput);
+
+            switch (command) {
```
phualiting: I think this would be a good place to use enums?

phualiting (commented):
Overall, I think the code mostly follows the coding standard and method names are easy to understand. It would be clearer if todo, deadline and event were separated into different classes.

PR #347, opened 2022-08-24
[Kevin Chang] iP

kevinchangjk:
# Carbon

> "For all x there exist y in life such that y is more brutal than x." - [Nguyen Vu Khang 💀](https://github.com/nguyenvukhang) 

Do you often feel *unmotivated* and *overwhelmed* by all your work? Well, not to worry.

Introducing `Carbon`, the omnipotent task manager.

## Capabilities

**Carbon** is perhaps the greatest innovation the human race bore witness to since the [fidget spinner](https://www.amazon.com/fidget-spinner/s?k=fidget+spinner).

- Keeping track of your tasks
- Keeping track of deadlines
- Storing and deleting as many tasks as you will ever need in a hundred lifetimes
- Greeting you
- Scolding you

## Setup

The installation and setup is easy. All you need to do is this:
1. Download the latest `jar` file from [here](https://github.com/kevinchangjk/ip/releases)
2. Run the `jar` file with the following command
    ```bash
    java -jar /path/to/jar/file
    ```
3. You're all ready to use `Carbon`! 🔥 

## Roadmap

- [X] Basic Task Manager Functions
- [ ] Generative Chatting Functionality
- [ ] ~~Conquer the World~~ World Peace

## Contributing

If you are interested to build **Carbon** alongside me, please feel free to contact me through pigeon mail. However, before you gather the courage to do so, do ensure that you are at least able to write code of this quality.

_Excerpt from Carbon source code_

```java
public class Carbon {
    /**
     * Initializes and runs an instance of Carbon.
     *
     * @param args Additional arguments.
     */
        Carbon shell = new Carbon();
        shell.run();
    }
}
```


src/main/java/carbon/Carbon.java
```diff
@@ -0,0 +1,58 @@
...
+        while (this.isRunning) {
+            try {
+                String input = this.ui.printIn();
+                String log = this.parser.process(input);
+                if (log != "") {
+                    this.ui.printOut(log);
+                } else {
+                    this.ui.exit();
+                    this.isRunning = false;
+                }
```
malwaregarry, line 42: would removing these lines from the `try-catch` to make it more obvious what error you are catching?

malwaregarry: would removing these lines from the `try-catch` to make it more obvious what error you are catching?

src/main/java/carbon/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final String WORKING_DIR = System.getProperty("user.dir");
+    private static final String SAVEFILE = Storage.WORKING_DIR + "/../../../store/tasks.txt";
+    private static final String SAVEFILE_DIR = Storage.WORKING_DIR + "/../../../store/";
+
+    /**
+     * Constructs an instance of Storage class.
+     *
+     * @return Storage object.
+     */
+    public Storage() {}
```
malwaregarry, line 30: perhaps omit the constructor and make the functions static?

src/main/java/carbon/TaskList.java
```diff
@@ -0,0 +1,234 @@
...
+     * @throws CarbonException  If there are invalid parameters.
+     */
+    public String validateAndMark(String input, boolean isDone) throws CarbonException {
+        int taskNumber;
+        int len = input.length();
+        if (isDone) {
+            int requiredLen = "mark ".length();
+            if (len <= requiredLen) {
+                CarbonException invalidParam = new InvalidParamException(input);
+                throw invalidParam;
```
malwaregarry: would it better to convert this to a guard clause to avoid nested if-else?

src/main/java/carbon/Ui.java
```diff
@@ -0,0 +1,120 @@
...
+        + "  ▄████████    ▄████████    ▄████████ ▀█████████▄   ▄██████▄  ███▄▄▄▄   \n"
+        + " ███    ███   ███    ███   ███    ███   ███    ███ ███    ███ ███▀▀▀██▄ \n"
+        + " ███    █▀    ███    ███   ███    ███   ███    ███ ███    ███ ███   ███ \n"
+        + " ███          ███    ███  ▄███▄▄▄▄██▀  ▄███▄▄▄██▀  ███    ███ ███   ███ \n"
+        + " ███        ▀███████████ ▀▀███▀▀▀▀▀   ▀▀███▀▀▀██▄  ███    ███ ███   ███ \n"
+        + " ███    █▄    ███    ███ ▀███████████   ███    ██▄ ███    ███ ███   ███ \n"
+        + " ███    ███   ███    ███   ███    ███   ███    ███ ███    ███ ███   ███ \n"
+        + " ████████▀    ███    █▀    ███    ███ ▄█████████▀   ▀██████▀   ▀█   █▀  \n"
+        + "                           ███    ███                                   \n";
+
```
malwaregarry: nice ascii logo

src/main/java/carbon/error/InvalidFlagException.java
```diff
@@ -0,0 +1,38 @@
...
+        super(input);
+        this.type = type;
+        if (type == Task.Type.DEADLINE) {
+            this.flag = "/by";
+        } else {
+            this.flag = "/at";
+        }
+    }
+
+    /** @inheritDoc */
```
malwaregarry: i think you need to wrap it with brackets for it to work.
```{@inheritDoc}```

**Review by malwaregarry — COMMENTED** [review]
Overall, I think your code is really good. Left only some minor comments.

File: src/main/java/carbon/Parser.java [hunk 6]
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Parses and processes user input text.
+     * This method is only for advanced commands that require parameters from the user.
+     *
+     * @param input User input text.
+     * @return Execution log.
+     * @throws CarbonException  If an error is encountered during processing or execution.
+     */
+    private String processAdvanced(String input) throws CarbonException {
+        String lowerCaseInput = input.toLowerCase();
```
e1010101, line 80: Is toLowerCase needed here? The input is always lowerCase from the process method above.

src/main/java/carbon/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+/**
+ * Stores data on tasks and acts as interface for storage.
+ * Used to save data and load data from a default filepath.
+ */
+public class Storage {
```
e1010101, line 15: Compact and clean parsing system, it's great!

src/main/java/carbon/TaskList.java
```diff
@@ -0,0 +1,234 @@
...
+    /**
+     * Updates if a specific task is done or not.
+     * Performs input validation before executing.
+     *
+     * @param input User text input.
+     * @param isDone Whether the task is done or not.
+     * @return Execution log.
+     * @throws CarbonException  If there are invalid parameters.
+     */
+    public String validateAndMark(String input, boolean isDone) throws CarbonException {
```
e1010101, line 69: Could this be moved to the Parser class? That way, input that is handled by the TaskList should not be malformed.

src/main/java/carbon/TaskList.java
```diff
@@ -0,0 +1,234 @@
...
+        int taskNumber;
+        int len = input.length();
+        if (isDone) {
+            int requiredLen = "mark ".length();
```
e1010101: You can use regex to validate input! To match "mark [digit]" for example, you could use "(mark)\s\d+".

src/main/java/carbon/TaskList.java
```diff
@@ -0,0 +1,234 @@
...
+        if (taskNumber < 1 || taskNumber > this.tasks.size()) {
+            CarbonException outOfBounds = new OutOfBoundsException(taskNumber, this.tasks.size());
+            throw outOfBounds;
+        } else {
+            String log = this.setTaskDoneness(taskNumber, isDone);
+            return log;
+        }
+    }
+
+    private String setTaskDoneness(int taskNumber, boolean isDone) {
```
e1010101: Nit: setTaskDoneness should be setDone according to the coding standard [here](https://se-education.org/guides/conventions/java/intermediate.html#naming).

src/main/java/carbon/Ui.java
```diff
@@ -0,0 +1,120 @@
...
+    // actual introduction
+    private static final String INTRO = "Hey, Carbon here. ";
+    private static final String[] INIT_PROMPTS = {
+        "What's up?",
+        "How's things going?",
+        "Nice weather today, huh?",
```
e1010101, line 16: Great personality!

src/main/java/carbon/Ui.java
```diff
@@ -0,0 +1,120 @@
...
+    // scanner for inputs
+    private Scanner sysScanner;
+
+    /**
+     * Constructs an instance of the Ui class.
+     *
+     * @return Ui object.
+     */
+    public Ui() {
+        this.rand = new Random();
```
e1010101, line 40: Creative implementation to randomise greetings/goodbyes!

src/main/java/carbon/task/Event.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final int TYPEKEY = Task.getTypeKey(Task.Type.EVENT);
+
+    private Temporal dateTime;
+
+    private Event(String name, Boolean isDone, Temporal dateTime) {
+        super(name, isDone);
+        this.dateTime = dateTime;
+    }
+
+    private static String extractName(String input) {
```
e1010101, line 28: The purpose of this method is a little unclear. If the input is valid, "name" will always be "event ", right? If this method is meant to check validity, perhaps it can also be moved to Parser?

src/main/java/carbon/task/Event.java
```diff
@@ -0,0 +1,133 @@
...
+    public String encode() {
+        int typeKey = Event.TYPEKEY;
+        int isDone = this.isDone ? 1 : 0;
+        String result = String.format(
+                "%d|%d|%s|%s\n", typeKey, isDone, this.name,
+                this.displayTime(Task.FormatType.READ));
+        return result;
+    }
+
+    private String displayTime(Task.FormatType formatType) {
```
e1010101, line 109: Nit: some common methods in Deadline and Event can be extracted to the base Task class to remove redundancy.

src/main/java/carbon/task/Task.java
```diff
@@ -0,0 +1,138 @@
...
+            decodedTask = Deadline.loadTask(name, isDone, param);
+            break;
+        case EVENT:
+            decodedTask = Event.loadTask(name, isDone, param);
+            break;
+        default:
+            // should never reach here
+            decodedTask = Todo.createTask("todo _");
+        }
+        decodedTask.changeDoneness(isDone);
```
e1010101: Since any Task is initialised with the isDone field parsed from input, this may be unnecessary.

e1010101 (commented):
Generally well-written and clean, with good organisation throughout. LGTM besides a few nits.

src/main/java/carbon/Carbon.java
```diff
@@ -0,0 +1,58 @@
+package carbon;
+
+import carbon.error.CarbonException;
+
+/**
+ * Interacts with user through CLI to keep track of tasks.
```
sltsheryl, line 6: Clearly written JavaDocs

src/main/java/carbon/Carbon.java
```diff
@@ -0,0 +1,58 @@
...
+ * Requires user to provide input and call task management commands.
+ */
+public class Carbon {
+    private boolean isRunning;
+    private Ui ui;
+    private Parser parser;
+    private Storage storage;
+
+    /**
+     * Creates an instance of Carbon.
```
sltsheryl, line 15: Could add that this is a constructor?

src/main/java/carbon/Carbon.java
```diff
@@ -0,0 +1,58 @@
...
+            } catch (CarbonException error) {
+                this.ui.printError(error);
+            }
+        }
+    }
+
+    /**
+     * Initializes and runs an instance of Carbon.
+     *
+     * @param args Additional arguments.
```
sltsheryl: It's not written exactly in the coding standard, but I think some places said it's not needed to have the @param tag for main method JavaDocs 
https://docs.oracle.com/javase/tutorial/getStarted/application/

src/main/java/carbon/Parser.java
```diff
@@ -0,0 +1,97 @@
+package carbon;
+
+import carbon.error.CarbonException;
+import carbon.error.InvalidInputException;
+import carbon.task.Task;
+
+/**
+ * Parses user inputs, and processes them.
```
sltsheryl, line 11: 👍🏻  Clear class documentation

src/main/java/carbon/Parser.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Parses and processes user input text.
+     * This method is only for advanced commands that require parameters from the user.
+     *
+     * @param input User input text.
+     * @return Execution log.
+     * @throws CarbonException  If an error is encountered during processing or execution.
+     */
+    private String processAdvanced(String input) throws CarbonException {
```
sltsheryl: Method name could be clearer like processAdvancedCommands

src/main/java/carbon/Storage.java
sltsheryl: I think it's okay to make it non-static. In the example code provided, they still did 
`storage = new Storage(filePath);`

src/main/java/carbon/Storage.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public Storage() {}
+
+    /**
+     * Loads the savefile and retrieves saved tasks.
+     * Uses the default filepath.
+     *
+     * @return TaskList instance containing the saved tasks.
+     */
+    public TaskList loadSavefile() {
```
sltsheryl: Could make it Camel case? Because Savefile could be the name of the file but loadSaveFile looks better

src/main/java/carbon/TaskList.java
sltsheryl: Second this.

src/main/java/carbon/TaskList.java
```diff
@@ -0,0 +1,234 @@
...
+        Task task = this.tasks.get(taskNumber - 1);
+        task.changeDoneness(isDone);
```
sltsheryl: changeDoneness can also have a formal name like changeCompleteStatus

src/main/java/carbon/Ui.java
```diff
@@ -0,0 +1,120 @@
...
+    // actual introduction
+    private static final String INTRO = "Hey, Carbon here. ";
+    private static final String[] INIT_PROMPTS = {
+        "What's up?",
+        "How's things going?",
+        "Nice weather today, huh?",
+        "How can I help you?",
+        "Please don't talk to me.",
+        "To get an A+ for CS2103T, you have t---[REDACTED]"
+    };
```
sltsheryl, line 20: Nice, this makes your bot unique with different replies :)

src/main/java/carbon/Ui.java
```diff
@@ -0,0 +1,120 @@
...
+        System.out.println(divider);
+        System.out.println("==> " + text);
+    }
+
+    /**
+     * Wraps any taking in of input with Carbon's text decorators.
+     *
+     * @return User input.
+     */
+    public String printIn() {
```
sltsheryl: Perhaps printInput() and printOutput() could sound clearer

src/main/java/carbon/error/CorruptedSavefileException.java
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * @param data String message for the error.
+     * @return CorruptedSavefileException object.
+     */
+    public CorruptedSavefileException(String data) {
+        super(data);
+        this.data = data;
+    }
+
+    /** @inheritDoc */
```
sltsheryl: I think it has to include curly braces
`{@inheritDoc}` but prof also said it's unnecessary in one of the forum posts if we don't want to write for overridden methods

src/test/java/carbon/ParserTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import carbon.error.InvalidInputException;
+
+public class ParserTest {
+    private Ui ui = new Ui();
+    private Storage storage = new Storage();
+    private Parser parser = new Parser(this.ui, this.storage);
+
+    @Test
+    public void process_invalidCommand_returnsInvalidInputException() {
```
sltsheryl, line 16: Clear unit testing naming 👍🏻

sltsheryl (commented):
An interesting bot with several extended and cool features 👍🏻 The JavaDocs were also written in a very complete and clear manner. Just take note about `{@inheritDoc}` and maybe some method names could be named more formally like doneness

src/main/java/carbon/Carbon.java
kevinchangjk: Yeah I think that would be better.

src/main/java/carbon/Parser.java
kevinchangjk: Fair point, I think I'll implement that. For now, I have this method name (`processAdvanced`) matching the previous method name (`process`), but `processCommand` and `processAdvancedCommand` are probably clearer.

src/main/java/carbon/Parser.java
kevinchangjk: Hmm yes, maybe. The `process` methods passes `input` not `lowerCaseInput` to `processAdvanced`, and the reason for this is that I can retain the original input string to be processed for task names. So I would be able to get stuff like "Review PR for iP" for my task names. Does that make sense to you, or maybe there's a simpler way?

src/main/java/carbon/Storage.java
kevinchangjk: True, I think making this purely static is a good option too. I'm not sure if there are any performance (or readability) differences between the two.

src/main/java/carbon/Storage.java
kevinchangjk: Oh yeah whoops, I have a habit of writing the term in a single word like "savefile". Thanks for catching that.

src/main/java/carbon/TaskList.java
kevinchangjk: I think that might be a good idea, I'll take a look into that.

src/main/java/carbon/TaskList.java
kevinchangjk: Oh that's a good suggestion, thanks! I'll try implementing that.

src/main/java/carbon/TaskList.java
kevinchangjk: Hmm yeah I think that's doable. I'll probably just need to remove the else statement and the braces.

src/main/java/carbon/TaskList.java
kevinchangjk: Ah okay, thanks for pointing that out.

src/main/java/carbon/TaskList.java
kevinchangjk: Actually shouldn't this probably follow the same format as above? Something like `setDone`?

src/main/java/carbon/Ui.java
kevinchangjk: Wow thank you. Not very useful with the GUI though.

src/main/java/carbon/task/Event.java
kevinchangjk: Maybe it's not very clear in code, but extractName uses the `String::substring` method to get the part of the user input after "event " until the flag (" /at"), essentially capturing the name that the user assigns to the task. 

I understand the point on checking validity though (same as the marking methods for `TaskList`). I currently have most of the validation decentralised among classes, so the `Parser` class mainly handles just the parsing and calling of processing methods. 

I do think having `Parser` perform validation makes sense too, though I'm not sure if it will result in overlapping input processing.

src/main/java/carbon/task/Event.java
kevinchangjk: Yes, I was thinking about this actually. Though `Todo` doesn't share the same complexities regarding dates and time, so maybe I'll create another abstract subclass of `Task` that has a `Temporal` field.

src/main/java/carbon/task/Task.java
kevinchangjk: Oh that's true, good catch.

src/main/java/carbon/task/Task.java
```diff
@@ -0,0 +1,138 @@
...
+        return decodedTask;
+    }
+
+    /**
+     * Updates whether the task is done or not.
+     *
+     * @param isDone Whether the task is done or not.
+     */
+    public void changeDoneness(boolean isDone) {
```
jgyj123: setter methods should follow the format set___. I think it will be better to call this method setDone.

src/main/java/carbon/Parser.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Parses and processes a user input text.
+     * If the command is not a basic command, this method will call the processAdvanced method.
+     *
+     * @param input User input text.
+     * @return Execution log.
+     * @throws CarbonException  If an error is encountered during processing or execution.
+     */
+    public String process(String input) throws CarbonException {
```
jgyj123: Great use of switch statements to improve the readability of the code. Well done!

src/main/java/carbon/task/Deadline.java
```diff
@@ -0,0 +1,133 @@
...
+            String timeString = input.substring(flagIndex + Deadline.FLAG.length() + 1);
+            try {
+                Temporal dateTime = Deadline.formatTime(timeString);
+                return dateTime;
+            } catch (DateTimeParseException error) {
+                CarbonException invalidTime = new InvalidTimeException(timeString);
+                throw invalidTime;
+            }
+        }
+    }
```
jgyj123: Perhaps it may be better to store the value of flagIndex + Deadline.FLAG.length() + 1 into a variable so that the code will be more readable.

src/main/java/carbon/Ui.java
jgyj123: I completely agree here, this gives your bot such a unique spin. Well done!

src/main/java/carbon/task/Todo.java
```diff
@@ -0,0 +1,67 @@
...
+public class Todo extends Task {
+    private static final int TYPEKEY = Task.getTypeKey(Task.Type.TODO);
+
+    private Todo(String name, Boolean isDone) {
+        super(name, isDone);
+    }
+
+    private static String extractName(String input) throws CarbonException {
+        int len = input.length();
+        int requiredLen = "todo ".length();
```
jgyj123, line 18: Just want to commend you on this point here and overall in the code - I really like how you make it a point not to use any magic numbers and not take shortcuts. Well done!

jgyj123 (commented):
Overall, LGTM. I really like that your code is very organised and that there are no extremely long methods, making your code very readable. Well done! 👍

PR #356, opened 2022-08-24
[Yang] ip

Shawn532:
# Duke
> Success is not an accident, success is actually a choice. -Stephen Curry

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download it from [here](https://github.com/Shawn532/ip/releases).
2. double-click it.
3. add your tasks.
4. let it manage yours tasks for you :+1: :yum:

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] GUI (coming soon)
- [ ] Reminders (coming soon)


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Initializes the Duke chatbot.
+     * @param filePath Specifies the path to save the content.
+     */
+
+    public Duke(String filePath) {
+        ui = new Ui();
+        tasks = new TaskList();
+        storage = new Storage(filePath);
+    }
```
brendantwh: Blank line between the Javadocs comment and the method declaration.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,33 @@
...
+     * @param filePath the path to save the file and load the file from
+     */
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    /**
+     * Updates the file
+     * @param str The strings to be written in the file
+     */
```
brendantwh: Nitpicking here, but there should be a blank line between description and the param section.

brendantwh (commented):
Couldn't find much coding standard violations, apart from the minor ones mentioned. Code seemed pretty clean and Javadocs for relevant methods were concise. Good job!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * start the program
+     */
+    public void run() {
+        ui.greetings();
+        while (true) {
+            String input = ui.readCommand();
+            String command = input.split(" ")[0];
+            switch (command) {
```
stevenlimhw: Perhaps, the commands may be better represented using enums to ensure that only commands included in the enum is considered as a valid command. Another idea would be to create classes for the specific types of commands.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,25 @@
...
+    public String eventTime;
+    public LocalDate exactTime;
+
+    public Event(String name, String eventTime) {
+        super(name);
+        this.eventTime = eventTime;
+        try {
+            this.exactTime = LocalDate.parse(eventTime);
+        } catch (DateTimeParseException e) {
+            System.out.println("Wrong input format!");
```
stevenlimhw: The error message can be made clearer here. Perhaps, you can show how the correct input format is really like.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Adds a task
+     *
+     * @param item The String that needs to be parsed and add to the list
+     * @throws DukeException Throws an exception
+     */
+    public void addATask(String item) throws DukeException {
+        String[] arr = item.split(" ", 2);
+        String type = arr[0];
+        switch (type) {
```
stevenlimhw: This is not a code quality violation, but rather a coding standard violation. In CS2103T, if I remember correctly, the cases should not have an additional indent with respect to the switch line.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,25 @@
+package duke;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+
+public class Deadline extends Task {
+    public String dl;
```
stevenlimhw: The naming of this field is quite unclear. Perhaps, you can try naming it with more descriptive words instead of just letters.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,25 @@
...
+    public LocalDate exactTime;
```
stevenlimhw: For the sake of information hiding (an OOP principle), fields of the class should be private. In the case when we need to access the field data, we can simply create a getter method that returns the data.

PR #366, opened 2022-08-24
[Shi Wen] iP

shwene:
## Duke 💯 

> “Everything has beauty, but not everyone can see.” — [Confucius](https://youtu.be/og1Pbn8OufI)

### Duke helps you store whichever tasks needs be. It's

- [x] Free
- [x] **Fun**
- [x] Fantastic

#### To use:
1. Register [here](https://youtu.be/Nrk8sqZfsgI)
2. Download the installer
3. *Run*

#### Disclaimer:
- The creator assumes no responsibility or liability for any errors or failures
- You agree and acknowledge that Duke is entitled to modify, improve or discontinue any of its services at its sole discretion

#### Here's a peak at the `main` method:

```java
}
```


src/main/java/duke/Command.java
```diff
@@ -0,0 +1,79 @@
...
+                    ui.updateTask(updatedTask, STATUS_COMPLETE);
+                } else if (keyword.equals("unmark")) {
+                    Task updatedTask = tasks.unmarkTask(index);
+                    ui.updateTask(updatedTask, STATUS_INCOMPLETE);
+                } else if (keyword.equals("delete")) {
+                    Task deletedTask = tasks.delete(index);
+                    ui.deleteTaskConfirmation(deletedTask, tasks.getSize());
+                }
+            }
+        }
```
beetee17: Would it be possible to clean this up and avoid nested if-else statements?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,51 @@
...
+        super(desc);
+        timing = getTiming(desc);
+    }
+
+    private String getTiming(String desc) {
+        int index = desc.indexOf('/');
+        if (index > 0) {
+            return desc.substring(index + 1);
+        }
+        return "OTOT";
```
beetee17: You might want to consider extracting this into a static final variable, instead of using it inline as a "magic value". This would also help you avoid future refactoring if used in multiple places.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,66 @@
...
+     * Displays to the user that there are no outstanding tasks.
+     */
+    protected void showLoadingError() {
+        System.out.println("No records detected, creating new file...");
+    }
+
+    /**
+     * Displays linebreak to users.
+     */
+    protected void lineBreak() {
```
beetee17: Nice job on following all the naming conventions throughout the PR! Just in this case you might want to use a verb to better specify the method's action e.g. `displayLineBreak`

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,104 @@
...
+        if (content.isBlank()) {
+            throw new DukeException(String.format("The description of a %s cannot be empty", keyword));
+        }
+    }
+
+    private static void isValidKeyword(String keyword) throws DukeException {
+        if (!isKeyword(keyword) || keyword.isBlank()) {
+            throw new DukeException("I'm sorry, but I don't know what that means :-(");
+        }
+    }
```
beetee17: I initially thought to suggest changing the return type to a boolean due to the method name. On second thought, this seems valid, but I would suggest renaming the methods to something like `checkIsValidTask` or `validateTaskDescriptionIsNotEmpty`

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,50 @@
...
+        while (s.hasNext()) {
+            String taskText = s.nextLine();
+            char taskType = taskText.charAt(3);
+            if (taskType == 'T') {
+                tasks.add(new Todo(taskText));
+            } else if (taskType == 'D') {
+                tasks.add(new Deadline(taskText));
+            } else if (taskType == 'E') {
+                tasks.add(new Event(taskText));
+            }
```
beetee17, line 40: Consider extracting task type identifiers into a class field, e.g. `Todo.TYPE_IDENTIFIER`

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    public static void main(String[] args) {
+        new Duke("data/tasks.txt").run();
+    }
+
+    private void run() {
+        ui.intro();
+        //int id = 0;
```
beetee17: I think you forgot to remove this commented-out piece of code!

beetee17 (commented):
Great job overall regarding code quality! Just a few issues here and there that are easily fixable. Keep it up!

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,82 @@
...
+    protected String enumerateList() {
+        ListIterator<Task> it = tasks.listIterator();
+        StringBuilder output = new StringBuilder();
+        while (it.hasNext()) {
+            output.append(it.nextIndex() + 1).append(" ").append(it.next()).append("\n");
+        }
+        return output.toString();
+    }
+
+    protected String find(String term) {
```
jeromepui: `find` seems a little vague. Perhaps you could rename it to `findTasks` instead.
```suggestion
    protected String findTasks(String term) {
```

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,82 @@
...
+        tasks.set(index, performedTask);
+        return performedTask;
+    }
+
+    /**
+     * Updates a Task at a given index in the collection as incomplete.
+     * @return Task updated as incomplete.
+     */
+    protected Task unmarkTask(int index) {
+        Task undidTask = tasks.get(index).undoTask();
```
jeromepui: `undidTask` doesn't seem to be standard English. Perhaps you could rename it to `incompleteTask`.
```suggestion
        Task incompleteTask = tasks.get(index).undoTask();
```

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,79 @@
...
+        storage.save(tasks.enumerateList());
+    }
+
+    /**
+     * Check whether a given Command instance is consists of the keyword bye (non case-sensitive).
+     * @return true if the command instance is a bye command, false otherwise.
+     */
+    protected boolean isBye(){
```
jeromepui: `isBye` seems vague. Perhaps you could rename it to `isByeCommand`.
```suggestion
    protected boolean isByeCommand(){
```

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Interacts with the console of Duke users, displaying operations, status, and errors.
+ */
+public class Ui {
+
+    /**
+     * Displays welcome message to users.
+     */
+    protected void intro() {
```
jeromepui: `intro` doesn't seem suitable for a method name. Perhaps you could rename it to `greet`.
```suggestion
    protected void greet() {
```

jeromepui (commented):
LGTM. 👍 Just need to fix some minor naming issues.

src/main/java/duke/Duke.java
shwene: Thanks for all the pointers!

src/main/java/duke/Ui.java
shwene: Thanks for all the pointers!

PR #372, opened 2022-08-24
[Tan Cheong Hsien Ryan] iP

ketamethane:
### Welcome to Duke

> “Sooner or later, those who win are those who think they can.” — Paul Tournier

Duke helps you keep track of your tasks! It is

- text-based
- quick to master
- ~~quick~~ efficient to use

All you need to do is to...
1. download it from [here](https://github.com/ketamethane/ip/blob/master/out/artifacts/iP_jar/iP.jar)
2. click on **View Raw**
3. add your tasks
4. let it manage your tasks for you 😊 

It is **FREE**! :)

Features:

- [x] Managing tasks
- [ ] managing deadlines (coming soon)
- [ ] Reminders (coming soon)

---

```
public class Main {
    }
}
```


src/main/java/seedu/duke/Deadline.java
```diff
@@ -0,0 +1,57 @@
...
+ * Represents a Deadline task.
+ */
+public class Deadline extends Task {
+
+    protected LocalDate dueDate;
+    protected String dueTime;
+
+    /**\
+     * Creates a Deadline task that has a date but no time included.
+     * @param description description of Deadline.
```
pwjj2000: Great and consistent work on the Javadoc documentation throughout your submission! 

Although, it might be better to leave a blank line between the description and the `@param` so that it is more aligned to the convention. Similar comments for the rest of the documentation as well.

Please note that you might have accidentally included an additional backslash `\` in line 14 as well.

src/main/java/seedu/duke/DukeException.java
```diff
@@ -0,0 +1,16 @@
...
+        super(message);
+    }
+
+
+
+//    public static void main(String[] args) {
+//        try {
+//
+//        }
+//    }
```
pwjj2000: Will it be more reader-friendly for other reviewers if commented out code is removed instead? Similar questions for other parts of the code

src/main/java/seedu/duke/Parser.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param isOpen ArrayList containing a boolean to check whether Duke is open or closed
+     * @param scanner Scanner object.
+     */
+    public static void parse(String command, String input, TaskList taskList, ArrayList<Boolean> isOpen, Scanner scanner) {
+
+        String[] inputArr = input.split(" ");
+//        String command = inputArr[0];
+        System.out.println("____________________________________________________");
+
+        switch (command) {
```
pwjj2000: Very neat work by using switch cases! 👍

src/main/java/seedu/duke/Storage.java
```diff
@@ -0,0 +1,194 @@
...
+                taskList.add(deadline);
+            }
+        } else {
+//            System.out.println("added E");
+            Event event = new Event(description, datetime);
+            if (isMarked.equals("1")) {
+                event.mark();
+            }
+            taskList.add(event);
+        }
```
pwjj2000: Have you considered using switch cases as well here since the conditionals are a little deeply nested?

src/main/java/seedu/duke/TaskList.java
```diff
@@ -0,0 +1,154 @@
...
+                String[] sliceByDesAndBy2 = desAndBy2.split(" /at ");
+                String description3 = sliceByDesAndBy2[0];
+                String dueTime2 = sliceByDesAndBy2[1];
+                Task event = new Event(description3, dueTime2);
+                taskList.add(event);
+                System.out.println("Got it. I've added this task:");
+                System.out.println("  " + event);
+                break;
+
+        }
```
pwjj2000: Could you double check on the indentation for this again? If I remember correctly, `case` should be aligned with `switch`, though I could be wrong as well.

pwjj2000 (commented):
Great piece of code with no major issues!

PR #388, opened 2022-08-24
[Marcus Pang Yu Yang] iP

marcuspang:
Duke🥚 frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ **_BLAZINGLY_** FAST to use

All you need to do is,

1. download it from [here](https://github.com/marcuspang/ip/releases/).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

<hr />

If you're a Java programmer, you can use it to practice Java too. Here's the `main` method:

```java
public class Main {
        new Duke("data/saveFile.txt").run();
    }
}
```


src/main/java/commands/Command.java
```diff
@@ -0,0 +1,28 @@
...
+     * @param storage The storage that is used to manage tasks saved locally.
+     * @throws DukeException if the command did not complete successfully.
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+
+    /**
+     * Indicates whether the chatbot should terminate after the command is executed.
+     *
+     * @return whether the chatbot should terminate.
+     */
```
ShummyOwnzYou, line 27: Perhaps the @return tag could be "returns true if the chatbot should terminate, false otherwise"

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Filters the task list for tasks that have a description that includes the specified query.
+     *
+     * @param query The specified query.
+     * @return New TaskList containing filtered tasks.
+     */
+    public TaskList filter(String query) {
+        return new TaskList(this.list.stream()
```
ShummyOwnzYou: nice usage of stream

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,143 @@
...
+    public void showMatchingTasks(TaskList tasks) {
+        System.out.println("     Here are the matching tasks in your list:");
+        this.showTasks(tasks);
+    }
+
+    /**
+     * Prints the goodbye message when the chatbot shuts down.
+     */
+    public void showGoodbye() {
+        System.out.println("     Bye. Hope to see you again soon!");
```
ShummyOwnzYou: Perhaps you could consider using an indentation constant to append in front of all the `System.out.println` commands

ShummyOwnzYou (commented):
Overall, the code follows the coding standard.
No glaring mistakes that I can see. Good job!

src/main/java/commands/Command.java
marcuspang: That's a good point, will update it

src/main/java/duke/Ui.java
marcuspang: That's a good point, but since I have removed the indentation for level 10, I won't be addressing this issue.

TheSoggy (commented):
Overall it LGTM! Coding standards were adhered to as far as I can see.

PR #389, opened 2022-08-24
[Fang Yiye] iP

yiyefyy:
# **Duke iP**

### A **_Bullet_** list
- todo
- deadline
- event
- - -
### A **_Numbered_** list
1. date
2. time
- - -
### A **_fenced_** code block
```ruby 
public void unmark(String input) {
        System.out.println("This task is marked as not done:");
        int j = Integer.parseInt(input.substring(7)) - 1;
        Task task = taskList.get(j);
        task.markAsNotDone();
        System.out.println(task);
    }
```
- - -
### A **_task_** list
- [X] eat 
- [ ] sleep
- - -
### An **_Emoji_**
👌
- - -
- - -
### A hyper link 
[ip](https://github.com/yiyefyy/ip)
- - -
### Inline code
This function `public void greet() {
        System.out.println("Hello!, I'm Yiye.\nWhat can I do for you? ◠‿◠");
    }` greets you. 
- - -


src/main/java/Duke/Commands.java
```diff
@@ -0,0 +1,152 @@
...
+        ui.listTask(taskList.listTasks());
+    }
+
+    /**
+     * Mark a task as done using the task index
+     *
+     * @param input read task index
+     */
+    public void markDone(String input) {
+        System.out.println("Nice! I have marked this task as done:");
```
aaronliwh: Should this be handled by the command or the UI class or the command class ?

src/main/java/Duke/Commands.java
```diff
@@ -0,0 +1,152 @@
+package Duke;
+import java.util.ArrayList;
+
+/**
+ * Contains methods of the program
+ */
+public class Commands {
+    private final TaskList taskList;
+    private final Ui ui;
```
aaronliwh: Would a single command class be any different from just adding the methods in Duke ?
Consider having an abstract Command class and having other commands extend from the Command class.

src/main/java/Duke/Parser.java
```diff
@@ -0,0 +1,56 @@
...
+        this.ui = ui;
+        this.commands = new Commands();
+    }
+
+    /**
+     *
+     * @param input that is being read
+     * @throws DukeException when input is invalid
+     */
+    public void parse(String input) throws DukeException{
```
aaronliwh: Should the error for invalid input (for example, unmark -100) be handled here ?
Also, consider using a switch statement instead

src/main/java/Duke/Storage.java
```diff
@@ -0,0 +1,76 @@
+package Duke;
+import java.io.*;
```
aaronliwh: Try not to use * imports, instead import what you need explicitly

aaronliwh (commented):
I think your code is generally good, just a few changes to the code design that you may want to consider.

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Constructor of Duke to initialise ui, storage and scanner
+     *
+     */
+    public Duke() {
+        ui = new Ui();
+        this.sc = new Scanner(System.in);
+        this.taskList = new TaskList();
+        filePath = "/Users/yiye/Desktop/cs2103Projects/ip/Data/duke.txt";
```
nqt230: Would it better to declare the file path as a static constant in the Storage class? I think that this makes sense because Storage is supposed to handle all the file reading and writing operations, so the Duke class does not need to know the file path.

src/main/java/Duke/Duke.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static void main(String[] args) throws DukeException {
+        Duke duke = new Duke();
+        Parser parser = new Parser(duke, ui);
+        duke.run(parser);
+    }
+
+
+    /**
+     * run the program
```
nqt230: I think the JavaDocs here could be more detailed in the parameter and exception descriptions, so for e.g.
`* @throws DukeException If the parser parses an invalid input.`

nqt230 (commented):
Looks good :+1:
Overall I think your code is easy to understand, with just a few concerns here and there regarding coding style and Javadocs, but nothing major.

PR #392, opened 2022-08-24
[Aung Khant Kyaw] iP

AkkFiros:


src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public String getDescription() {
+        return this.description;
+    }
+
+    /**
+     * Obtains and returns the completion status of the task
+     * @return The completion status
+     */
+    public boolean getCompletion() {
```
jonasgwt: Perhaps this method name should be phrased to sound more like a boolean for example, `isCompleted`

src/main/java/ToDo.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Constructor for a ToDo task
+     * @param description description of a task that the user inputs
+     */
+    public ToDo(String description) {
+        super(description);
+    }
+
+    /**
+     * toString method for a ToDo task
```
jonasgwt: Perhaps we can start our method header comments in the form `Returns ...`, `Sends ...`, `Adds ...`

src/main/java/Deadline.java
```diff
@@ -0,0 +1,33 @@
+/**
+ * Deadlines are tasks that have a description and must be done by a certain time.
+ */
+public class Deadline extends Task {
+    protected String by;
+
+    /**
+     * Constructor for Deadline task
+     * @param description description of a task that the user inputs
```
jonasgwt: It would be better to have a full stop at the end of each parameter description.

src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
...
+        return this.isDone;
+    }
+
+    /**
+     * Obtains and returns the icon representation of the completion status of the task
+     * @return the icon representing the completion status of the task
```
jonasgwt: I think it would be better if we start the description of our `@return` with a capital letter.

src/main/java/KKBot.java
```diff
@@ -0,0 +1,168 @@
...
+                        String description = task.getDescription();
+                        String index = String.valueOf(i + 1);
+                        //if last task in list, add a divider after
+                        if (i == tasks.size() - 1) {
+                            System.out.println(index + ". " + task.toString() + "\n" + divider);
+                        } else {
+                            System.out.println(index + ". " + task.toString() + "\n");
+                        }
+                    }
+                }
```
maxtance: I feel that the code here is deeply nested. You could consider creating different methods to handle different things. For instance, a UI class to handle the printing and a getList method to get the current list of tasks.

src/main/java/KKBot.java
```diff
@@ -0,0 +1,168 @@
...
+                            System.out.println(divider);
+                        } else {
+                            System.out.println("Silly you, that task is already complete!\n" + divider);
+                        }
+                    }
+                } catch (NumberFormatException nfe) {
+                    System.out.println("Woopsies you left out the task number. Please specify one!\n" + divider);
+                } catch (IndexOutOfBoundsException ioobe) {
+                    System.out.println("Hmm... that task doesn't exist. Please input another task number!\n" + divider);
+                }
```
maxtance: I like the use of try-and-catch blocks. Great job on considering and handling the different exceptions.

src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
+/**
+ * Tasks are added by the user for KKBot to store and keep track of.
+ *
+ * @author AkkFiros
+ */
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
maxtance: Good use of access modifiers which follows OOP 👍

src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
+/**
```
maxtance: You could consider putting Task, ToDo, Deadline, and Event into a package.

maxtance (commented):
Overall, the basic level for code quality is met. To improve the code quality, I feel that there could be more abstractions, especially in the main method. Keep it up 👍

PR #399, opened 2022-08-24
[Pratham Jain] iP

pratham31012002:
# Pluto
> "The key is not to prioritise what's on your schedule, but to schedule your priorities." - Stephen Covey ([source](https://www.forbes.com/sites/kevinkruse/2012/07/16/the-7-habits/?sh=3e0f4d839c6d))

**Pluto** frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/pratham31012002/ip/releases/download/A-Jar/pluto.jar).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you.

And it is **FREE!**
Features:

- [x] Managing tasks :+1:
- [x] Managing Deadlines :+1:
- [x] Adding reminders for events :+1:

```
public class Main {
    }
}
```


src/main/java/pluto/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+            case DELETE:
+                return new DeleteCommand(parseIdx(textArr[1]));
+            case LIST:
+                return new ListCommand();
+            case SHOW:
+                return new ShowCommand(parseDateOnly(textArr[1]));
+            case BYE:
+                return new ExitCommand();
+            default:
+                throw new PlutoException("\tOOPS!!! I'm sorry, but I don't know what that means :-(");
```
Nephelite: I didn't like this indentation of ```case``` clauses in your ```switch``` statement.

According to the module's coding standards, ```case``` clauses share the same indentation as your ```switch``` statement.

I will only mention this here, but please do change all the switch statements to which this applies, such as the one below this block.

src/main/java/pluto/Parser.java
```diff
@@ -0,0 +1,115 @@
+package pluto;
+
+import pluto.command.*;
```
Nephelite: I didn't like the use of wildcard import. 

Even if you are using all files under ```pluto.command```, perhaps consider explicitly importing everything to comply with module coding conventions?

src/main/java/pluto/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    public static int parseIdx(String idx) throws PlutoException {
+        try {
+            return Integer.parseInt(idx.strip()) - 1;
+        } catch (NumberFormatException e) {
+            throw new PlutoException("\tOOPS!!! Index of task is required.");
+        }
+    }
```
Nephelite: I like that you put everything related to parsing into this Parser class. Good use of Abstraction!

src/main/java/pluto/Parser.java
```diff
@@ -0,0 +1,115 @@
...
+        EVENT,
+        MARK,
+        UNMARK,
+        DELETE,
+        LIST,
+        SHOW,
+        BYE
+    }
+
+    public static Command parse(String inputLine) throws PlutoException {
```
Nephelite, line 50: I didn't like how undescriptive this method name is. You have other methods with which you use to parse other inputs. 

Perhaps consider renaming this method to ```parseCommand``` instead?

src/main/java/pluto/TaskList.java
```diff
@@ -0,0 +1,72 @@
...
+                missions.get(idx).markAsDone();
+            } else {
+                missions.get(idx).markAsUndone();
+            }
+        } catch (IndexOutOfBoundsException e) {
+            throw new PlutoException("\tOOPS!!! Valid index required.");
+        }
+    }
+
+    public int nTasks() {
```
Nephelite, line 116: I didn't like this method name, because it is a public method with an undescriptive name. 

Perhaps a more descriptive method name would be better.

src/main/java/pluto/Ui.java
```diff
@@ -0,0 +1,35 @@
+package pluto;
+
+import java.util.Scanner;
+
+public class Ui {
+    private static final String CHATBOT = "pluto.Pluto";
+    private Scanner sc;
```
Nephelite: I didn't like this Scanner object declaration's placement. Perhaps consider moving it below or above all the constants? It's not a coding style violation, I just have OCD.

src/main/java/pluto/command/DeleteCommand.java
```diff
@@ -0,0 +1,33 @@
...
+        try {
+            storage.rewriteFile(tasks);
+            StringBuilder deleteMessage = new StringBuilder();
+            deleteMessage.append("\tNoted. I've removed this task:\n");
+            deleteMessage.append(String.format("\t\t%s\n", t.toString()));
+            deleteMessage.append(String.format("\tNow you have %d tasks in the list.", tasks.nTasks()));
+            ui.print(deleteMessage);
+        } catch (IOException e) {
+            tasks.addTask(idx - 1, t);
+            throw new PlutoException("\tOOPS!!! Couldn't delete task. Try again!");
```
Nephelite: I didn't like this solution for deleting a task from TaskList. 

Consider the scenario where your try block throws an Exception that is not IOException. Perhaps it throws an unchecked exception. In this scenario, you would lose the task to be added back into the TaskList because the catch block is not executed. This is a prime example of using Exceptions as control flow, which is something you should avoid.

Perhaps consider separating the file rewriting process from the task deletion process. The TaskList failing to be stored or rewritten should not mean that the delete command fails entirely.

src/main/java/pluto/command/UpdateStatusCommand.java
```diff
@@ -0,0 +1,38 @@
...
+            } else {
+                markMessage.append("\tOK, I've marked this task as not done yet:\n");
+            }
+            markMessage.append("\t\t" + tasks.getTask(idx).toString());
+            ui.print(markMessage);
+        } catch (IOException e) {
+            tasks.markTask(idx, !isDone);
+            throw new PlutoException("\tOOPS!!! Couldn't update task status. Try again!");
+        }
+    }
```
Nephelite: I like that you combined both Mark and Unmark command into a single command. It is a very efficient use of code. Really clean.

Unfortunately, this code block still shares the issues I had mentioned in DeleteCommand. Refer to my comments there for more details.

src/main/java/pluto/task/Deadline.java
```diff
@@ -0,0 +1,31 @@
+package pluto.task;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    private LocalDateTime by;
```
Nephelite, line 12: I feel mixed about the name of this variable. It is technically enough, but I still feel like it should be more descriptive. Could you consider making it more descriptive?

(This also applies to the similar private variable in Event)

src/main/java/pluto/task/Task.java
```diff
@@ -0,0 +1,46 @@
...
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+
+
+    public abstract String toFile();
```
Nephelite, line 59: I really like this method that you used to turn tasks into storage. If given the time, I would want to try to do the same as well. The idea is really solid.

My only concern is an ignorant or malicious user. What if they run a command like 
todo try | have to complete project

Nephelite (changes requested):
Overall I think your project is really clean. and has a lot of great ideas and implementations of OOP concepts. I really liked the combination of MarkCommand and UnmarkCommand into a single UpdateStatusCommand.

However, it makes a few errors in using exceptions for control flow and is also slightly too idealistic regarding the user giving bad inputs. 

Also, there are a few module coding convention violations with wildcard imports and the switch case clauses.

pratham31012002:
> Overall I think your project is really clean. and has a lot of great ideas and implementations of OOP concepts. I really liked the combination of MarkCommand and UnmarkCommand into a single UpdateStatusCommand.
> 
> However, it makes a few errors in using exceptions for control flow and is also slightly too idealistic regarding the user giving bad inputs.
> 
> Also, there are a few module coding convention violations with wildcard imports and the switch case clauses.

Thank you for your deep insights and valuable feedback. I will fix these issues!

PR #400, opened 2022-08-24
[Anjali Agarwal] iP

agarwal-anjali:
### Alpha
_“Time isn’t the main thing. It’s the only thing.”_ - Miles Davis [(source)](https://quotefancy.com/miles-davis-quotes)

Never miss a deadline again! 
**Alpha** is here to help you keep a track of your tasks! 🫡

It's 
- ~EASY~ SUPER EASY to learn
- ~FUN~ SUPER FUN to use

To begin with, all you need to do is:

1. Download it from [here](https://github.com/agarwal-anjali/ip/releases/download/A-Jar/alpha.jar) and double-click on it.
2. Here you go, Alpha is now all yours and that too for **free**!
3. Just enter **_help_** to learn about all commands and their format.
4. That's it, you can now start managing your tasks!

It provides a range of features:

- [x] Add tasks: todos, events and deadlines.
- [x] Mark and unmark tasks as done/pending.
- [x] Find all tasks that contain a specific keyword.
- [x] Delete tasks.
- [x] View the list of tasks.
- [x] Get help.

If you are a Java enthusiast and would like to know about the range of colors Alpha uses to make its UI attractive then, here's the `ANSI_CODE` for all the colors:

    /** Ansi code to add white font color to print statements */
    public final String ANSI_WHITE = "\u001B[37m";

    /** Ansi code to add red font color to print statements */
    public final String ANSI_RED = "\u001B[31m";

    /** Ansi code to add bold blue font color to print statements */
    public final String ANSI_BLUE_BOLD = "\033[1;36m";

    /** Ansi code to add blue font color to print statements */
    public final String ANSI_BLUE = "\u001B[36m";

    /** Ansi code to reset default font color in print statements */
    public final String ANSI_RESET = "\u001B[0m";


src/main/java/Alpha.java
```diff
@@ -0,0 +1,146 @@
...
+        catch (Exception e){
+            System.out.println(e.getMessage());
+        }
+    }
+
+    private static void welcomeMessage() {
+        System.out.println("\n-----------------\n" + ANSI_BLUE + "Hello, I'm ALPHA!" + ANSI_RESET + "\n-----------------");
+    }
+
+    private static void enterMessage(Scanner in) throws AlphaException, IOException {
```
Echomo-Xinyu: `enterMessage` seems to suggest a function that enters a message into something. Do you think `readMessage` might be a better name as it reflects the action from the chatbot's point of view instead of user's point of view?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,23 @@
...
+    static final String ANSI_RED = "\u001B[31m";
+    static final String ANSI_RESET = "\u001B[0m";
+    protected String deadline;
+
+    DateTimeFormatter dTF = DateTimeFormatter.ofPattern("MMM d yyyy");
+    public Deadline(String description, String deadline, String taskType) {
+        super(description, taskType);
+        this.deadline = deadline;
+    }
+    public String getDeadline(){
```
Echomo-Xinyu: Do you think `getDeadlineTimeString()` can be a better time as we would expect a `String` instead of a `Deadline` object?

Similar questions apply for `Event` and `Todo` class.

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
+public abstract class Task {
+    public static final String ANSI_YELLOW = "\u001B[33m";
+    public static final String ANSI_RESET = "\u001B[0m";
+    protected String description;
+    protected boolean status;
+
+    protected String taskType;
```
Echomo-Xinyu: I think an empty line should be left between the `class attributes` and the `constructor`?

src/main/java/Alpha.java
```diff
@@ -0,0 +1,146 @@
+import java.io.FileNotFoundException;
```
Echomo-Xinyu: Missing package header as required.

Echomo-Xinyu (commented):
It can be seen that much effort has been put into writing this work and most of the function namings and variable namings are following the required conventions.

Considering the progress of the writer so far (till Level-8 as declared in the tag), he or she might need to put in more time into writing this project.

But still, good job for the decent code quality!

Hongyi6328 (commented):
Overall, good attempt! A lot of requirements fulfilled.
Just some minor suggestions here.
Would you consider adding more header comments before non-private classes and methods, so that future reviewers can understand your project better?
And don't forget to add a PR description according to the link here! [link](https://nus-cs2103-ay2223s1.github.io/website/schedule/week4/project.html)

PR #402, opened 2022-08-24
[Zizheng] iP

dr-arrgghh:


src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+    public void chatDuke() throws DukeException {
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> taskArray = new ArrayList<>();
+        String str = sc.nextLine();
+        int counter = 0;
+        if (str.equals("bye")) {
+            System.out.println("Bye. Hope to see you again soon");
+        }
+        while (!str.equals("bye")){
+            if (str.equals("list")) {
```
Prearo: Consider using a switch statement here to reduce clutter and reduce nesting of if-else clauses.

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
+import java.util.*;
```
Prearo: Try to avoid wildcard imports and try to add some header comments for your classes and methods.

src/main/java/Events.java
```diff
@@ -0,0 +1,13 @@
+public class Events extends Task{
```
Prearo: Does this represent one event or multiple events?

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+
+            }
+
+            else {
+                taskArray.add(new Task(str));
+                counter++;
+                System.out.println("added: " + str);
+                str = sc.nextLine();
+            }
+
```
Prearo: Some whitespace is good, but try to avoid too many empty lines. I find it hard to follow.

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+            } else {
+                return;
+            }
+        }
+    }
+
+    public void checkInput(String str) throws DukeException {
+        if (str.equals("")) {
+            throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+        }
```
Prearo: Try to stick to K&R style bracketing - else if should be on the same line as the closing } of the previous if clause.
Moreover, since each result is a thrown exception, there is no need for 'else'.
Also, what is 'blah' here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+            }
+
+            else if (str.equals("mark")) {
+                int num = sc.nextInt();
+                Task currTask = taskArray.get(num - 1);
+                currTask.mark();
+                System.out.println("Nice! I've marked this task as done:\n" +
+                        "[" + currTask.getStatusIcon() + "] " + currTask.description);
+                str = sc.nextLine();
+                str = sc.nextLine();
```
Prearo: Is there a reason for the double nextLine()? Is it due to the input format?

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+            counter++;
+            str = sc.nextLine();
+        }
+
+
+
+
+    }
+
+    public void returnArrayStrings(ArrayList<Task> tskArray){
```
Prearo: I think naming the parameter taskArray here is at least clearer.

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println("Hello from\n" + logo);*/
     }
+
+    public void chatDuke() throws DukeException {
+        Scanner sc = new Scanner(System.in);
+        ArrayList<Task> taskArray = new ArrayList<>();
+        String str = sc.nextLine();
+        int counter = 0;
```
Prearo: Consider using ArrayList.size() to replace an explicit counter variable to get the last task in the list.

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+
+
+
+
+    }
+
+    public void returnArrayStrings(ArrayList<Task> tskArray){
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < tskArray.size(); i++){
+            if (tskArray.get(i) != null) {
```
sikai00: Nitpick here, but might be good to avoid using ```tsk.Array.get(i)``` twice. Instead, assign its return value to a variable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+                System.out.println(Integer.toString(i + 1) + "." + tskArray.get(i));
+            } else {
+                return;
```
sikai00: Given that the for loop runs for tskArray.size(), is it really necessary to try and break with ```return;```?

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+                        taskArray.get(counter));
+                counter++;
+                System.out.println("Now you have " + counter + " tasks in the list.");
+                str = sc.nextLine();
+            }
+            else if (str.equals("deadline")) {
+                String input = sc.nextLine();
+                checkInput(input);
+                String by = sc.nextLine();
+                checkInput(by);
```
sikai00: ```checkInput```'s exception message does not seem to match the argument being used (not a todo?). More importantly, there seems like better and cleaner ways to check the validity of inputs.

sikai00: It also seems like there is no indicator for the user to know what to enter at which point, since your arguments are all separated by a newline. Maybe some feedback would be good for UX?

sikai00 (commented):
Most of the comments I had were already pointed out @Prearo. There's a lot of room for improvement in my opinion, in terms of cleaner code. More effort and work should also be put in for the styling of code.

PR #411, opened 2022-08-24
[Linus Lee] iP

zupey:
# Duke v0.1
Use Duke to keep track of tasks without worrying about forgetting them!

It's
* text-based
* efficient
* ~~cheap~~**FREE**

Just follow these 3 easy steps:
1. download it from [here](https://github.com/zupey/ip/releases)
2. run it.
3. win!

Features:
- [x] Manage tasks
- [ ] Remember deadlines
- [ ] Track events
- [ ] Seamless experience

The `faster` you type, the `better` Duke is. 💯 

Sample commands:
```bash
todo Make Pull Request
bye
```

 ## Hear from our users
> This is the most life-changing application I have used - John Doe


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+ * Duke is a program that helps its user to store tasks.
+ */
+public class Duke {
+
+    private void initialise() {
+        wrapWithLines("Hello from\n" + logo);
+
+        File storageDirectory = new File("./data");
+        if (!storageDirectory.exists()) {
+            if (!storageDirectory.mkdir()) {
```
shaniceng: Perhaps you could use this method to make your if else statement not nested for better readability :)
```suggestion
            if (!storageDirectory.mkdir() && !storageDirectory.mkdir()) {
```

src/main/java/duke/commands/CommandEventHandler.java
```diff
@@ -0,0 +1,46 @@
...
+    public CommandEventHandler(String value, String flag, String additionalValue) throws DukeException {
+        super(value, flag, additionalValue);
+    }
+
+    /**
+     * Checks the validity of the syntax after being parsed by CommandParser.
+     * @throws DukeException error message.
+     */
+    public void checkValid() throws DukeException {
+        if (value == null || !flag.equals("at") || additionalValue == null) {
```
shaniceng: Perhaps you could try to avoid complicated expressions here by initialising them first and use them in the if else statement
```suggestion
        boolean isValid = value == null || !flag.equals("at") || additionalValue == null;
        if (isValid) {
```

src/main/java/duke/tasks/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Constructor of taskList linked to a Storage.
+     * @param db the database where taskList stores the tasks.
+     */
+    public TaskList(Storage db) {
+        this.taskList = new ArrayList<>();
+        this.db = db;
+    }
+
+    public String getRemainingTasks() {
```
shaniceng, line 43: Great job in naming all the methods in verbs and camel cases 👍

src/main/java/duke/commands/Parser.java
```diff
@@ -0,0 +1,40 @@
...
+        return command;
+    }
+
+    public List<String> parseCommand(String command) {
+        Matcher matcher = Parser.commandParser.matcher(command);
+        matcher.find();
+        String value = matcher.group(2);
+        String flag = matcher.group(3);
+        String additionalValue = matcher.group(4);
+        List<String> result = new ArrayList<>();
```
shaniceng: Should this be "results" instead since it is a list and the name of variable should be plural
```suggestion
        List<String> results = new ArrayList<>();
```

shaniceng (commented):
Overall, all of your codes are easy to read and it LGTM 👍 
Just some nits to be fixed :)
Good Job!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        Storage db = new Storage("./data/duke.txt");
+        TaskList taskList = db.load();
+
+        Scanner sc = new Scanner(System.in);
+        CommandHandlerFactory commandHandlerFactory = new CommandHandlerFactory();
+        String command;
+        while (!(command = sc.nextLine()).equals("bye")) {
```
ngquyduc: Might it be a better idea to let the Parse do all the parsing commands work?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,57 @@
...
+import duke.exceptions.DukeException;
+import duke.tasks.TaskList;
+
+
+/**
+ * Duke is a program that helps its user to store tasks.
+ */
+public class Duke {
+
+    private void initialise() {
```
ngquyduc: Perhaps abstracting this method in a class _Storage_ might be a good idea.

src/main/java/duke/commands/CommandDeadlineHandler.java
```diff
@@ -0,0 +1,44 @@
...
+        super(value, flag, additionalValue);
+    }
+
+    /**
+     * Checks the validity of the syntax after being parsed by CommandParser.
+     * @throws DukeException error message.
+     */
+    public void checkValid() throws DukeException {
+        if (value == null || !flag.equals("by") || additionalValue == null) {
+            throw new DukeException("Correct usage: deadline return book /by 24/04/2019 1600");
```
ngquyduc: I like the way you throw an exception and help the user to type in the correct command at the same time.

src/test/java/duke/tasks/DeadlineTest.java
```diff
@@ -0,0 +1,19 @@
+package duke.tasks;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+    @Test
+    public void dbRepresentationFromUserInput() {
```
ngquyduc:
```suggestion
    public void getDbRepresentationFromUserInput() {
```

ngquyduc (commented):
Overall, clean and easy-to-follow code 👍 .
However, you can consider some of my suggestions.

src/main/java/duke/Duke.java
zupey: Thank you for your suggestion!

src/main/java/duke/commands/Parser.java
zupey: Thank you for pointing this out!

src/main/java/duke/commands/CommandEventHandler.java
zupey: Thanks for the suggestion!

PR #414, opened 2022-08-24
[Daniel Lee] iP

lulucopter:
# Dan

Looking at your repository, this application sparked some ideas on adding in new features to spice up your program. I have hence created `Dan`, a **sp00ky** :ghost:  variation of your `Duke` program.

Dan is an interactive, *boosted* to-do list, planner all in one, freeing up your headspace so you can keep your focus on the now without worry of forgetting anything important.

> "Having a clear mind and a clear space allows you to think and act with purpose" - Erika Oppenheimer ([source](https://www.erikaoppenheimer.com/blog/two-must-dos-for-better-organization))

The application is 
- easy to use
- easy to learn
- multiplies your productivity
- And the best part, ***SUPER FAST***

To get started, just follow the steps below:
1. Download it from [here](https://github.com/lulucopter/ip/releases)
2. Go to the directory which you had installed it in
3. Open the application using `java -jar dan.jar` (make sure you have java on your computer!)
4. And you are ready to get started! 😄 

There are the commands that are currently available:
- [x] `bye` (exits the program)
- [x] `list` (lists out the tasks that are currently in your list)
- [x] `mark <task number>` (mark a task as completed)
- [x] `unmark <task number>` (unmark a completed task)
- [x] `todo <task description>` (creates a to-do task)
- [x] `deadline <task description> /by <datetime>` (creates a deadline)
- [x] `event <task description> /at <datetime>` (creates an event)
- [x] `delete <task number>` (deletes a task)
- [x] `find <keyword>` (finds a task)

Here is the `main` method in case you can't find it
```java
        Application.launch(Dan.class, args);
}
```


src/main/java/Dan/Dan.java
```diff
@@ -0,0 +1,53 @@
+package Dan;
+
+import java.io.IOException;
+import java.nio.file.NoSuchFileException;
+
+import java.util.Scanner;
+
+
+public class Dan {
```
Donovan9617: Perhaps you forgot java docs for this class and its methods?

src/main/java/Dan/Parser.java
```diff
@@ -0,0 +1,57 @@
...
+            case "todo":
+                //fall through
+            case "deadline":
+                //fall through
+            case "event":
+                tasks.addTask(input);
+                break;
+
+            default:
+                throw new DanException("I don't really understand what do you mean by that...");
```
Donovan9617: Nice job following the coding standards for the switch clause!

src/main/java/Dan/TaskList.java
```diff
@@ -0,0 +1,103 @@
...
+public class TaskList {
+    private List<Task> tasks;
+
+    public TaskList(List<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    public void addTask(String input) throws DanException {
+        String description;
+        String dateString;
```
Donovan9617: Did you forget access modifiers here?

src/main/java/Dan/Task.java
```diff
@@ -0,0 +1,28 @@
...
+        isDone = done;
+    }
+
+    public String toDataString(String separator) {
+        return String.format("%s%d%s%s", separator, (this.isDone) ? 1 : 0, separator, this.description);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[%s] %s",getStatusIcon(), this.description);
```
Donovan9617: There should be a space gap between , and getStatusIcon()

Donovan9617 (commented):
Nice job following coding standards. Overall LGTM!

src/main/java/Dan/Parser.java
```diff
@@ -0,0 +1,57 @@
...
+    public Parser(TaskList tasks) {
+        this.tasks = tasks;
+    }
+    
+    public boolean parse(String input) {
+        String action = input.split(" ")[0];
+        try {
+            switch (action) {
+            case "bye":
+                Ui.sayonara();
```
zsiggg: Sayonara isn't in English, so maybe we could change this to `Ui.bye()` instead?

src/main/java/Dan/Parser.java
```diff
@@ -0,0 +1,57 @@
...
+            case "list":
+                tasks.showTasks();
+                break;
+
+            case "mark":
+                tasks.markTask(Integer.parseInt(input.split(" ")[1]));
+                break;
+
+            case "unmark":
+                tasks.unMarkTask(Integer.parseInt(input.split(" ")[1]));
```
zsiggg: Since method names should be in camelCase, perhaps this should be `tasks.unmarkTask()` instead? Because I see unmark as one word.

src/main/java/Dan/Parser.java
zsiggg: It's good that the case clauses are not indented, but I think there shouldn't be an empty line between cases as well? Maybe its for clarity, but from what I read in the style guide it seems like there shouldn't be the extra lines.

src/main/java/Dan/TaskList.java
```diff
@@ -0,0 +1,103 @@
...
+        if (input.startsWith("todo")) {
+            description = input.replace("todo", "").strip();
+            if (description.isEmpty()) {
+                throw new DanException("Please provide me a description for your todo item");
+            }
+            tasks.add(new ToDo(description));
+
+        } else if (input.startsWith("deadline")) {
+            String[] temp = input.replace("deadline","").strip().split("/by");
```
zsiggg: Because we are using `temp` in multiple places within this `else if` block, might want to name it more informatively? Perhaps like `inputArr`?

src/main/java/Dan/TaskListReader.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class TaskListReader {
+    /* Stores data separated by <SEPARATOR> in the format with no whitespace:
+    eg. SEPARATOR = "|"
+    Tasktype{T,D,E}|Task.isDone{1,0}|Task.Name|(optional)Deadline String
+     */
```
zsiggg: Maybe these could be made as JavaDoc comments for `TaskListReader`?

src/main/java/Dan/TaskListReader.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+        return taskList;
+    }
+
+    public void writeTaskListToFile(TaskList taskList) throws IOException {
+        Files.write(filePath, taskList.getTasks().stream()
+                .map(x -> x.toDataString(SEPARATOR))
+                .collect(Collectors.toList())
+        );
```
zsiggg: Nice use of streams 🔥

zsiggg (commented):
Overall, coding standard mostly adhered to. LGTM after the few comments are resolved, and as the previous reviewer mentioned, the JavaDocs for public classes/methods are made. :)

src/main/java/Dan/Dan.java
lulucopter: Thanks for spotting this!

src/main/java/Dan/Parser.java
lulucopter: Thanks for pointing this out! Will take note of this.

src/main/java/Dan/TaskList.java
lulucopter: These are actually part of addTask method, and not an attribute of the class

src/main/java/Dan/Task.java
lulucopter: Missed this out. Thank you!

src/main/java/Dan/Parser.java
lulucopter: Had forgotten about this! Thanks for pointing it out

src/main/java/Dan/Parser.java
lulucopter: You are right. Thought that unMark were 2 different words

src/main/java/Dan/TaskList.java
lulucopter: Thanks for the suggestion

PR #418, opened 2022-08-24
[Song Ivan] iP

songivan00:
# **Duke Bot**
> “If You Fail to Plan, You Are Planning to Fail” — Benjamin Franklin. [source](https://www.goodreads.com/quotes/460142-if-you-fail-to-plan-you-are-planning-to-fail#:~:text=Quote%20by%20Benjamin%20Franklin%3A%20%E2%80%9CIf,you%20are%20planning%20to%20fail!%E2%80%9D)

### Duke is an interactive _to-do list_ that is:
- Text-based
- Easy to use
- ~Fast~SUPER FAST

### A few commands you need to know:
1. _list_: see your to do list
2. _mark **task index**_: to mark the task as completed
3. _todo **task**_: to add a task to the to do list
4. _deadline **task** /by **the timing of the deadline**_: to add a task with a deadline to the to do list
5. _event **task** /at **the timing of the event**_: to add a task with a event timing
6. _delete **task index**_: to delete the task 
7. _bye_: to close the bot

### With this bot:
- [x] No more procrastination 🥳
- [x] No more late submissions 🥳
- [x] No more forgetting events 🥳

If you are a Java programmer, you can use it to practice Java too. Here's the main method:
```ruby  
    new Duke("./src/main/java/duke.txt").run();
}
```


src/main/java/duke/Event.java
```diff
@@ -0,0 +1,40 @@
+package duke;
+
+/**
+ * Task with a period of time of the task
+ */
+public class Event extends Task{
+    /**
+     * Event object at field which indicates the timings
+     */
+    protected String at;
```
Guanzhou03: I'm curious as to whether the access modifier can be private or what goes into the decision to make it protected

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,240 @@
...
+            } else if (first.length() >= 5 && first.substring(0, 5).equals("event")) {
+                eventMessage(first, lst, store);
+                return false;
+            } else if (first.length() == 8 && first.substring(0, 6).equals("delete")) {
+                deleteMessage(first, lst, store);
+                return false;
+            } else if (first.length() >= 4 && first.substring(0, 4).equals("find")) {
+                findMessage(first, lst, store);
+                return false;
+            }
```
Guanzhou03: LGTM, can consider using switch case to reduce the number of lines

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,240 @@
...
+     * Marks a task as completed
+     *
+     * @param str The users input
+     * @param lst The tasklist
+     * @param storage the storage object
+     */
+    public static void markMessage(String str, TaskList lst, Storage storage) throws DukeException, IOException {
+        char index = str.charAt(5);
+        int number = Integer.parseInt(String.valueOf(index));
+        if(number > lst.count()) {
```
Guanzhou03: Good to have a space after the "if"

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    /**
+     * Finds a task based on a keyword
+     *
+     * @param str The users input
+     * @param lst The tasklist
+     * @param storage the storage object
+     */
```
Guanzhou03: Good that all public methods have JavaDoc comments

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,47 @@
...
+
+
+/**
+ * Task with a date to complete the task by
+ */
+public class Deadline extends Task {
+    /**
+     * Deadline object by field which indicates the deadline
+     */
+    protected LocalDate by;
```
onepersonhere: Perhaps a name should explain the named entity to the reader accurately and at a sufficient level of detail?
```suggestion
    protected LocalDate byDate;
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,55 @@
+package duke;
+
+import java.io.FileNotFoundException;
+import java.util.*;
```
onepersonhere: Perhaps imported classes should always be listed explicitly?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /**
+     * Adds a Todo to the tasklist
+     *
+     * @param str The users input
+     * @param lst The tasklist
+     * @param storage the storage object
+     */
+    public static void todoMessage(String str, TaskList lst, Storage storage) throws DukeException, IOException {
```
onepersonhere: Perhaps use verb for actions?
```suggestion
    public static void parseTodoMessage(String str, TaskList lst, Storage storage) throws DukeException, IOException {
```

onepersonhere: I found that many of the method names have the same issue.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public ArrayList<Task> load() throws FileNotFoundException, IOException {
+        ArrayList<Task> lst = new ArrayList<>();
+        File file = new File("./src/main/java/duke.txt");
+        Scanner sc = new Scanner(file);
+        while (sc.hasNextLine()) {
+            String currentLine = sc.nextLine();
+            String[] stringDetails = currentLine.split("\\|");
+            String taskType = stringDetails[0];
+            String isMark = stringDetails[1];
```
onepersonhere: Perhaps name it in such a way to not sound like a boolean variable?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,105 @@
...
+        t.unmark();
+        System.out.println("OK, I've marked this task as not done yet:\n  [ ] " + t.description);
+    }
+
+    /**
+     * A method that adds a task into the tasklist
+     *
+     * @param tsk The task
+     */
+    public void add(Task tsk) {
```
onepersonhere: Perhaps name should explain the named entity to the reader accurately and at a sufficient level of detail?
```suggestion
    public void addTask(Task tsk) {
```

onepersonhere: I found many similar methods with names of insufficient details.

onepersonhere (commented):
Overall, I found your code easy to read for the most part except for the huge else-if chunk. 
I noted a few minor coding standard & naming violations too. 
Consider renaming some methods/variables.

src/main/java/duke/Parser.java
songivan00: Thanks for the suggestions will implement that.

src/main/java/duke/Parser.java
songivan00: Thanks!

src/main/java/duke/Deadline.java
songivan00: Youre right, thank you!

src/main/java/duke/Parser.java
songivan00: thanks for pointing that out! will change them :)

src/main/java/duke/Storage.java
songivan00: ok! thank you!

PR #421, opened 2022-08-24
[Silas Tay] iP

SilasTSL:
# Luffy

> Relax, allow the mind to become empty, and surprise yourself with the great treasure that begins to flow from your soul. -Someone cool

Luffy helps you keep track of your tasks in a quick and neat manner. It includes:
- Text-based input
- Simple to pick up
- Super fast to use

All you need to do to use it is:
1. Download it [Here](https://github.com/SilasTSL/ip)
2. Double click it
3. Add your tasks
4. Enjoy!

Here's the `main` method for Luffy:
```
    /**
     * Main method for Luffy.
     *
     * @param args User commandline input
     */
        new Luffy(FILE_PATH).run();
    }
```

Checklist of things to do with Luffy:
- [x] Download Luffy
- [ ] Try adding Todos, Events and Deadlines
- [ ] Deleting Todos, Events and Deadlines

Best thing is, Luffy is **FREE**!

Enjoy :+1:!


src/main/java/luffy/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Constructor for Deadline task.
+     *
+     * @param description Description of Deadline
+     * @param by Deadline for deadline task
+     */
+    public Deadline(String description, String by) {
+        super(description);
+        this.type = "[D]";
+        this.by = by;
```
desmondyst, line 21: Any reason why you decided to put the `by` attribute as part of `Task` class? Perhaps it will be better to have it as an attribute of `Deadline` class as it is only applicable to `Deadline` class and not other class that inherit from `Task`. Same goes for other fields such as `at`.

src/main/java/luffy/Luffy.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param filePath File path to data file
+     */
+    public Luffy(String filePath) {
+        this.ui = new Ui();
+        this.storage = new Storage(filePath);
+        try {
+            this.tasks = new TaskList(storage.load());
+        } catch (DukeException e) {
+            this.ui.showLoadingError();
+            this.tasks = new TaskList();
```
desmondyst, line 51: Perhaps can come up with alternative to handle the logic here? In general, not the best practice to use `try-catch` block for control flow.

src/main/java/luffy/Storage.java
```diff
@@ -0,0 +1,102 @@
...
+        File dataFile = new File(this.filePath);
+        try {
+            if (dataFile.exists()) {
+                Scanner scanner = new Scanner(dataFile);
+                while (scanner.hasNext()) {
+                    Task currentTask;
+                    String[] taskLine = scanner.nextLine().split(Pattern.quote(" | "));
+                    String taskDescription = taskLine[2];
+                    switch (taskLine[0]) {
+                        case "[T]":
```
desmondyst: Just some small points on styling.
```suggestion
                    case "[T]":
```

src/main/java/luffy/Task.java
```diff
@@ -0,0 +1,48 @@
+package luffy;
+
+/**
+ * Task abstract class to represent a task. Tasks consist of Deadlines, Events and Todos.
+ *
+ * @author Silas Tay A0233425M
+ */
+public abstract class Task {
+    protected String description;
+    protected boolean isDone;
```
desmondyst, line 12: Good naming on boolean variable!

src/main/java/luffy/Task.java
```diff
@@ -0,0 +1,48 @@
...
+    protected String type;
+    protected String by;
+    protected String at;
```
desmondyst: Should this be extracted out into the relevant classes that `extends Task`? 🤔

src/main/java/luffy/Deadline.java
```diff
@@ -0,0 +1,32 @@
+package luffy;
+
+/**
+ * Deadline class to represent a deadline.
+ *
+ * @author Silas Tay A0233425M
+ */
+public class Deadline extends Task{
```
desmondyst:
```suggestion
public class Deadline extends Task {
```

src/main/java/luffy/Parser.java
```diff
@@ -0,0 +1,118 @@
...
+        this.ui = new Ui();
+    }
+
+    /**
+     * Parses through user inputs and takes action based on command.
+     *
+     * @param s User input line
+     * @param tasks Tasklist of Luffy instance
+     */
+    public void parse(String s, TaskList tasks) {
```
desmondyst: This `parse` method seems to form the main bulk of the logic that handles the user's command. Perhaps you can think of creating `Command` classes and extract the logic into each corresponding `Command` class. E.g. `DeadlineCommand` class.

src/main/java/luffy/Storage.java
desmondyst: Perhaps u can try to make use of `ENUMS`? As quoted from the textbook - "Use enumerations when a certain variable can take only a small number of finite values."

desmondyst (commented):
Majority of the code LGTM. Just a few nits to fix. However, the logic of the program seems to be nested in conditionals, which can get messy. Perhaps, you can try to separate the logic into smaller classes (e.g. the logic in `Parser` class). Other than that, good effort! 👍

PR #423, opened 2022-08-25
[Shan Hashir] iP

sh4nH:
### Duke Pro

>" Everything begins with an idea" - Earl Nightingale

Duke chat-bot is your pal for keeping track of your to-do list. It is 

- text-based
-  ~EASY~ SUPER EASY to learn and 
- super fast to use

Duke Pro is **free to use** too!
To start using Duke pro today, all you have to do is

1. download it from [here](https://github.com/sh4nH/ip)
2. double click on Duke Pro
3. Add your tasks
4. Let Duke manage your tasks for you 😉

Features of Duke Pro:

- [ ] Manage your tasks (todos,events and deadlines)
- [ ] Reminders (coming soon)
- [ ] Share your tasks with others (coming soon)

____________________________________________________________________________________________________________________________________

If you are interested to learn Java, it can help with that too. Here is the `main` method from DukePro:
```ruby
  }

### CODE REVIEW

File: src/main/java/Deadline.java [hunk 1]
```diff
@@ -0,0 +1,12 @@
+public class Deadline extends Task{
+    protected String by;
+    public Deadline(String taskName,String by) {
+        super(taskName);
+        this.by = by;
+    }
```
marcuspang: Your classes and methods require a description as per coding standards. And since you haven't done the `A-JavaDoc` level yet, I won't add a comment for methods/classes missing javadoc.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String taskName;
```
marcuspang:
```suggestion
    protected final String taskName;
```
Since `taskName` is only updated once (i.e. in your constructor), I feel it is good practice to add the `final` keyword. But, I don't think this is part of the coding standards so feel free to close this comment

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+    protected boolean isDone;
+    public Task(String taskName) {
+        this.taskName = taskName;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
```
marcuspang: I like this abstraction for the status icon as it allows for easy modification in the future 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,116 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
+    public  Duke(){
+
+    }
+    ArrayList<Task> tasks = new ArrayList<>();
```
marcuspang:
```suggestion
    public Duke() {
    }

    ArrayList<Task> tasks = new ArrayList<>();
```
You have some extra spacing in your constructor function declaration

src/main/java/Duke.java
```diff
@@ -1,10 +1,116 @@
...
+    private final static String horLine = "------------------------------------------------------";
+    private final static String welcomeMsg = horLine + "\nHello! I'm Duke\n" +
```
marcuspang: > Constant names must be all uppercase using underscore to separate words.

Perhaps you can name them something more semantic as well, like `HORIZONTAL_LINE` and `WELCOME_MESSAGE`

src/main/java/Duke.java
```diff
@@ -1,10 +1,116 @@
...
+            "What can I do for you?\n" + horLine;
+
+    private void addToList(String taskName) {
+        try{
+            //if task is a todo
+            if (taskName.matches("\\btodo\\s.*\\b")){
+                Task newTask = new ToDo(taskName.substring(5));
```
marcuspang: I feel that instead of `newTask` you can name it something else to indicate that it's a `ToDo` class instead of a normal `Task` object (perhaps newToDo). However, this is not part of the coding standards, so feel free to close this

src/main/java/Duke.java
```diff
@@ -1,10 +1,116 @@
...
+            else if (taskName.matches("\\bevent\\s.*\\s/at\\s.*\\b")){
+                String des = taskName.substring(6,taskName.indexOf("/")-1);
+                String at = taskName.substring(taskName.indexOf("/")+4,taskName.length());
+                Task newTask = new Event(des,at);
+                tasks.add(newTask);
+                System.out.println(horLine + "\n\tGot it. I've added this task:\n" + "\t" + newTask.toString());
+                System.out.println("\tNow you have " + tasks.size() + " tasks in the list." + "\n" + horLine);
+
+            }
+            ////if task is a deadline
```
marcuspang:
```suggestion
            //if task is a deadline
```
I think you might have added additional `/` by accident

src/main/java/Duke.java
```diff
@@ -1,10 +1,116 @@
...
+        } catch(Exception e) {
+            System.out.println(horLine + "\n\t" + e.getMessage() + "\n" + horLine);
+        }
+
+
+
+
+    }
+
+
```
marcuspang:
```suggestion
    }

```
>  Logical units within a block should be separated by one blank line.

Although this may technically not be considered a block, these empty lines are unnecessary. There are other instances of this, so I won't comment those to avoid spam

src/main/java/Duke.java
```diff
@@ -1,10 +1,116 @@
...
+        System.out.println(horLine + "\n\tNoted, I have removed this task: \n" +
+                "\t\t" + tasks.get(num-1).toString() + "\n" + "\tNow you have " +
+                tasks.size() + " tasks in the list." + "\n" + horLine);
+    }
+
+    private void viewList() {
+        System.out.println(horLine + "\n");
+        System.out.println("\tHere are the tasks in your list:\n");
+        for(int i = 0; i<tasks.size(); i++) {
+            int num = i+1;
```
marcuspang:
```suggestion
        for(int i = 0; i < tasks.size(); i++) {
            int num = i + 1;
```
>  White space within a statement

There's quite a lot of instances where whitespace is missing in your statements, which can be solved easily with an IDE. So do a run through your files to make sure they are formatted correctly.

marcuspang (commented):
There's quite a few coding standard violations, but they can be easily fixed with an IDE. The other issues will be resolved once you reach the `A-JavaDoc` level as well. 

Overall, I think these issues are quite minor, just that there are quite a few of them, and I like the fact that you abstracted the user commands into separate methods. 👍

src/main/java/Deadline.java
sh4nH: Hi, I will be adding the description soon. Thank you for reminding me to do it.

src/main/java/Duke.java
sh4nH: Yes, you are correct. I will do that in my next commit.

src/main/java/Duke.java
sh4nH: oh yes, thank you for pointing it out.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,128 @@
...
+                        ui.tasksLoadedMsg(tasks.size())), dukeImage)
+        );
+
+        sendButton.setOnMouseClicked((event) -> {
+            handleUserInput();
+        });
+
+        userInput.setOnAction((event) -> {
+            handleUserInput();
+        });
```
WingKei67: It would be better if arrow-head style could be avoided.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Add a task (to-do,event or deadline) to the tasklist.
+     *
+     * @param taskName description of the task to be added.
+     * @param ui       ui to print the confirmation messages after task is added
+     *                 to the list.
+     */
+    public String addToList(String taskName, Ui ui) {
```
WingKei67, line 44: This method may be a bit too long? 🤔

PR #424, opened 2022-08-25
[Tan Tze Yeong] iP

tantzeyeong:
# Duke

Duke frees your mind and lets you focus on the actual tasks! It's,

- text-based
- simple
- ~~fast~~ too fast! Works at *LIGHTSPEED* ⚡

All you got to do is,

1. download it from [here](https://github.com/tantzeyeong/ip/releases).
2. double-click it.
3. add your tasks.
4. let Duke remember them for you! 😄 

And guess what? It's completely **FREE**!! 😱 

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

If you're a Java programmer, you can practice Java with it too! Here's the `main` method:

```java
    new Duke("data/Duke.txt").run();
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
+import duke.command.Command;
+import duke.model.*;
```
dlimyy: Wildcard(aka *) should be avoided here so the classes should be stated explicitly.

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,37 @@
...
+
+import duke.model.TaskList;
+import duke.storage.Storage;
+import duke.ui.Ui;
+
+/**
+ * The command to be executed with respect to the user's input.
+ */
+public abstract class Command {
+    private Boolean isExit = false;
```
dlimyy, line 10: Good naming for Boolean variables!

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+        String[] input = userInput.split(" ", 2);
+
+        String inputKeyword = input[0];
+
+        Keyword keyword = Keyword.getKeyword(inputKeyword);
+
+        String[] description;
+
+        switch (keyword) {
+            case BYE:
```
dlimyy: There should not be any indentation for the cases clauses, so the word 'case' should have the same indentation as switch

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,94 @@
...
+    public static Command parse(String userInput) {
+        String[] input = userInput.split(" ", 2);
+
+        String inputKeyword = input[0];
+
+        Keyword keyword = Keyword.getKeyword(inputKeyword);
+
+        String[] description;
+
+        switch (keyword) {
```
dlimyy, line 87: Good use of switch cases to handle different commands!

dlimyy (commented):
Overall, the code quality is good and good coding practices are well adhered to!

src/main/java/Duke.java
tantzeyeong: Oh thank you for pointing that out!

PR #427, opened 2022-08-25
[Conrad Lew] IP

ConradLew:
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
2. double-click it.
3. add your tasks.

And it is **FREE!**

Features:
- [ ] Managing tasks
- [ ] Managing deadlines
- [ ] Reminders
---

```Java
public class Main {
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+public class Duke {
+
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+
+    /**
+     * Constructor for the <code>Duke</code> chatbot.
```
zsiggg, line 18: Not sure if there is an error here, for \<code>Duke\</code>?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Returns the error message arising from issues in loading or storing tasks locally.
+     *
+     * @return String containing the error message.
+     */
+    public String showLoadingError(Exception e) {
+        return e.getMessage();
+    }
+
```
zsiggg, line 29: `showLoadingError` has the same implementation as `showError`. Perhaps could consider using just one method out of the two?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,87 @@
...
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.loadTasks());
+        } catch (DukeException e) {
+            ui.showError(e);
+            tasks = new TaskList();
+        } catch (IOException e) {
+            ui.showLoadingError(e);
+        }
```
zsiggg, line 32: I'm assuming that `IOException` would be thrown from `storage.loadTasks()`, due to an error when reading the output file. Will there be a need to reinitialise `tasks`, since this error would mean that there are no initial tasks to be shown? 

To illustrate,
```java
} catch (IOException e) {
    ui.showLoadingError(e);
    tasks = new TaskList();
}
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,87 @@
...
+            case DEADLINE:
+                this.tasks.addTasks(Parser.parseDeadlineTask(userResponse));
+                return this.ui.showAddTaskSuccess(this.tasks);
+            case DELETE:
+                this.tasks.deleteTask(Parser.parseTaskNumber(userResponse));
+                return this.ui.showRemoveTaskSuccess(Parser.parseTaskNumber(userResponse), this.tasks);
+            case FIND:
+                return this.ui.showMatchingTasks(Parser.parseSearchInput(userResponse), this.tasks);
+            default:
+                break;
```
zsiggg: Would there be an error message shown if the user gives an invalid command?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.Arrays;
+
+/**
+ * Represents a parser for user input given to Duke chatbot.
+ *
+ * @author Conrad
+ */
+public class Parser {
+
```
zsiggg, line 13: Perhaps we should make `Parser` abstract, since it only contains static methods?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+     * Parses the user input into a <code>Todo</code> task.
+     *
+     * @param userInput String input provided by user.
+     * @return An <code>Todo</code> task with the specified description.
+     */
+    public static Todo parseTodoTask(String userInput) {
+        String[] parsedUserResponse = userInput.split(" ");
+        String newTaskDescription = String.join(" ",
+                Arrays.copyOfRange(parsedUserResponse, 1, parsedUserResponse.length));
+        return new Todo(newTaskDescription);
```
zsiggg, line 42: Maybe we could do something like this, assuming we are trying to obtain the todo description from the command given?
```java
String[] newTaskDescription = userInput.replace("/todo", "").strip();
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,172 @@
...
+                Arrays.copyOfRange(parsedUserResponse,
+                        1,
+                        atSeparationIndex));
+        String newEventTime = String.join(" ",
+                Arrays.copyOfRange(parsedUserResponse,
+                        atSeparationIndex + 1,
+                        parsedUserResponse.length));
+        return new Event(newTaskDescription, newEventTime);
+    }
+
```
zsiggg, line 65: Using the same idea as the comment on `parseTodoTask`, and assuming we are trying to get the description (which is before /at) and the time (which is after /at) maybe we could do this? If you agree it is easier to understand.
```java
String[] parsedUserResponse = userInput.replace("/event").split("/at");
String newTaskDescription = parsedUserResponse[0];
String newEventTime = parsedUserResponse[1];
```

zsiggg (commented):
Really amazing that you used JavaFX to create a GUI 😃. Just got a few things that I feel could make the code cleaner :)

src/test/java/duke/DeadlineTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.time.LocalDate;
+
+import org.junit.jupiter.api.Test;
+
+public class DeadlineTest {
+
+    @Test
+    public void testGetTextRepresentation() {
+        Deadline testDeadline = new Deadline("test", LocalDate.parse("1999-02-02"));
```
Kennethsim2000, line 13: Indentation for wrapped line should be 8 spaces

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Returns the tasks matching the user query string.
+     *
+     * @param searchInput The query string to search for.
+     * @param tasks The current <code>TaskList</code>.
+     * @return String containing the matching tasks.
+     */
+    public String showMatchingTasks(String searchInput, TaskList tasks) {
+        String matchingStart = "Here are the matching tasks in your list:\n";
+        return matchingStart + tasks.getMatchingTasksRepresentation(searchInput);
```
Kennethsim2000, line 117: indentation for wrapped lines should be 8 spaces. E.g.
setText("Long line split"
        + "into two parts.");

Kennethsim2000 (commented):
Overally, good job with the code. LGTM:)

PR #431, opened 2022-08-25
[Jonathan Lam] iP

JonLamy:
# Duke
> Nobody wants tasks. But since we don't have a choice, why not rely on Duke for some help?

Duke is a task manager with features:
- Text-based
- _Easy_ to learn
- **Foolproof**
- and only ~~$1.99~~ **FREE**!

Download it [here](https://github.com/JonLamy/ip/releases) 🥳

Features:
- [x] Task management
- [ ] Motivation (soon)
---
If you love Java, here's a peek at Duke's `main` method:
```
public class Main {
    }
}
```

Have fun!


src/main/java/ip/Parser.java
```diff
@@ -0,0 +1,75 @@
...
+    /** String that the user has entered */
+    private Scanner inputLine;
+
+    /**
+     * Stores user input string as a Scanner object.
+     *
+     * @param s Scanner to be stored.
+     */
+    public void load(Scanner s) {
+        inputLine = s;
```
maxng17: Might be better to have it as Scanner inputLine so this line written as inputLine = inputLine

src/main/java/ip/Ui.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Scanner;
+
+/**
+ * Handles all communications with the user.
+ *
+ * @author Jonathan Lam
+ */
+public class Ui {
+    private Scanner input = new Scanner(System.in);
+    private final String greeting = "  ___          ___  \n"
```
maxng17: Following the 2103T code standards, greeting should be all caps since its a constant.

src/main/java/ip/Ui.java
```diff
@@ -0,0 +1,59 @@
...
+                                   + " (o o)        (o o) \n"
+                                   + "(  V  ) DUKE (  V  )\n"
+                                   + "--m-m----------m-m--";
+    private final String bye = " ^ ^         \n"
```
maxng17: All caps for bye like above.

src/main/java/ip/exception/BadDeadline.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public BadDeadline(String s) {
+        deadline = s;
+    }
+
+    @Override
+    public String toString() {
+        if (deadline.isEmpty()) {
+            return "Deadline not specified for event!\n"
+                   + "Correct example: `deadline Homework /by 2/5/2022 2359`";
```
maxng17: Might be better if the example is changed to DD/MM/YYYY 2359 or 31/8/2022 since the current example now, I can interpret it as DD/MM... or MM/DD making it abit confusing

src/main/java/ip/exception/BadDeadline.java
```diff
@@ -0,0 +1,23 @@
...
+        } else {
+            return "Deadline specified: \"" + deadline + "\" has incorrect formatting.\n"
+                   + "Correct example: `deadline Homework /by 2/5/2022 2359`";
```
maxng17: Same idea here as above

src/main/java/ip/exception/DukeException.java
```diff
@@ -0,0 +1,11 @@
+package ip.exception;
+
+/**
+ * Exceptions thrown by Duke.
+ */
+public class DukeException extends Exception {
+
+    public DukeException() {
```
maxng17, line 8: No real need for this constructor right just like the IndexNotSpecified, unless I am wrong

maxng17 (commented):
Overall pretty good. Just some nitpicking from me cause I really couldn't find much styling errors haha. 👍

src/main/java/ip/exception/BadDeadline.java
chanyuk: I agree! Or maybe another way would be to directly give instructions on formatting the date to users?

src/main/java/ip/exception/BadTimespan.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    @Override
+    public String toString() {
+        if (timespan.isEmpty()) {
+            return "Timespan not specified for event!\n"
+                   + "Correct example: `event Meeting /at Tuesday, 2-3pm`";
+        } else {
+            return "Timespan specified: \"" + timespan + "\" has incorrect formatting.\n"
+                   + "Correct example: `event Meeting /at Tuesday, 2-3pm`";
```
chanyuk: Maybe you can try including a more generalised example for formatting as suggested in your BadDeadline.java file?

chanyuk (commented):
Overall it was good! The error messages was clear and your classes seem well-abstracted.

PR #432, opened 2022-08-25
[Chrysline Lim] iP

ChryslineLim:
### DukePro

> “Your mind is for having ideas, not holding them.” – David Allen [(source)](https://dansilvestre.com/productivity-quotes)

- text-based
- easy to learn

All you need to do is, 

1. download it from [here](https://github.com/nus-cs2103-AY2223S1/ip/pull/432).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon) 

---

If you are a Java programmer, you can use it to practise Java too. Here's the `main` method:

```java
public class Main {
    }
}
```


src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,61 @@
+package duke.command;
+
+import duke.DukeException;
+import duke.TaskList;
+import duke.task.DeadlineTask;
+import duke.task.EventTask;
+import duke.task.Task;
+import duke.task.ToDoTask;
+
+import java.time.LocalDate;
```
shenyih0ng: Standard `java.` imports should be placed above `duke.` imports according to `checkstyle` rules

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,56 @@
...
+        
+        if (command.equals("bye")) {
+            return new ExitCommand();
+        } else if (command.equals("list")) {
+            return new ListCommand();
+        }
+        
+        if (!command.equals("mark") && !command.equals("unmark") && !command.equals("todo") && !command.equals("deadline") && !command.equals("event") && !command.equals("delete") && !command.equals("find")) {
+            throw new DukeException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+        }
```
shenyih0ng: Instead of doing a check for unknown commands here, maybe it will be clearer if you shift this integrate this check into the switch statement below (L33-L55).

For example:

```java
switch (command) {
.
.
.
default:
    throw new DukeException("...");
}
```

This will remove the need to write long condition checks as the number of command type increase.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,56 @@
...
+            return new AddCommand("todo", suffix);
+        case "deadline":
+            return new AddCommand("deadline", suffix);
+        case "event":
+            return new AddCommand("event", suffix); 
+        case "find": 
+            return new FindCommand(suffix);
+        default: {
+            int pos = Integer.parseInt(suffix) - 1;
+            return new DeleteCommand(pos);
```
shenyih0ng: Awesome job following the correct style for switch case statements 👍

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+import java.time.format.DateTimeFormatterBuilder;
+import java.util.ArrayList;
+import java.util.Locale;
+import java.util.Scanner;
+
+
+public class Storage {
+    private final String FILE_PATH;
+    
+    public Storage(String FILE_PATH) {
```
shenyih0ng: Great job at setting `FILE_PATH` as `final`!

However, I am not quite sure if `FILE_PATH` in this case can be considered a constant since you are initializing it in the constructor. Perhaps the naming convention in the case should not be that of a constant.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+    
+    /**
+     * Retrieves information from the data file to populate an ArrayList
+     *
+     * @return ArrayList of Tasks represented in the data file. 
+     * @throws DukeException If data file cannot be accessed.
+     */
+    public ArrayList<Task> load() throws DukeException {
+        ArrayList<Task> data = new ArrayList<>();
```
shenyih0ng: Instead of naming it as `data`, I would think it will be semantically clearer if it is named something like `tasks`, since it is meant to represent a collection of `Task`.

src/test/java/duke/command/TaskListStub.java
```diff
@@ -0,0 +1,27 @@
...
+
+import duke.DukeException;
+import duke.Storage;
+import duke.TaskList;
+import duke.UI;
+import duke.task.Task;
+
+import java.util.ArrayList;
+
+public class TaskListStub extends TaskList {
```
shenyih0ng, line 14: Great use of a stub for `TaskList` testing 👍

build.gradle
```diff
@@ -0,0 +1,41 @@
...
+        showExceptions true
+        exceptionFormat "full"
+        showCauses true
+        showStackTraces true
+        showStandardStreams = false
+    }
+}
+
+application {
+    mainClassName = "seedu.duke.Duke"
```
shenyih0ng: cmiiw, i believe the `mainClassName` should be `duke.Duke` instead since the your `Duke.java` is under the `duke` package

shenyih0ng (commented):
Overall the coding standards have larged been adhered to! There is also clear documentation and sufficient unit tests :) 

Perhaps you can also consider adding documentation for the remaining methods/classes in the future!

src/main/java/duke/Parser.java
ChryslineLim: I placed it on top in order to be before the "OOPS!!! The description cannot be empty." error, otherwise it would print that for any unknown commands of only one word :( I also couldn't shift the empty error to the end as I needed to ensure the commands have a description before the switch cases :(

build.gradle
ChryslineLim: Thank you for the review! I will make the necessary changes :)

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,25 @@
+package duke.task;
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
YZTangent, line 5: Great job naming booleans in the _isSomething_ format!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,104 @@
...
+                    break;
+                case 'D':
+                    String[] parts = s.substring(7).split(" \\(by: ");
+                    DateTimeFormatter pattern = new DateTimeFormatterBuilder()
+                            // case-insensitive to parse JAN and FEB
+                            .parseCaseInsensitive()
+                            // add pattern
+                            .appendPattern("MMM dd yyyy")
+                            // create formatter (use English Locale to parse month names)
+                            .toFormatter(Locale.ENGLISH);
```
YZTangent: Great job using higher level line breaks by breaking at every operation called on DateTimeFormatterBuilder!

src/main/java/duke/task/Task.java
YZTangent, line 8: Great job naming booleans in the isSomething format!

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import duke.command.AddCommand;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ParserTest {
+    
+    @Test
+    public void parse_todoBorrowBook_newAddCommand() throws DukeException {
```
YZTangent: Great job naming test methods using underscores and following the three part format!

YZTangent (commented):
The PR looks great! :) Good job following most of the coding conventions, such as casing and naming methods with verbs!

src/test/java/duke/ParserTest.java
ChryslineLim: Thank you for the review! :)

PR #442, opened 2022-08-25
[Jason Qiu] iP

jasonqiu212:
# Cheese

> Dogs are not our _whole_ life, but they make our lives _whole_. 

Cheese is a 🐶-themed task manager! It's...
- Command-line based (for the tech-savvy ones out there)
- **Simple** to use
- ~~Stressful~~ Stress-free to use with puppy phrases

### Features

- [x] Manage tasks
- [x] Manage deadlines and events
- [ ] Graphical user interface

## Set-up Guide
1. Download the latest [release](https://github.com/jasonqiu212/ip/releases/tag/A-Jar).
2. Click on it.
3. Add your tasks right away!

## Developer's Guide

For budding Java programmers, feel free to practice your skills on this repository! The entry point is at `Cheese.java`.
```java
    new Cheese("data/cheese.txt").run();
}
```


src/main/java/cheese/task/Task.java
```diff
@@ -0,0 +1,72 @@
...
+        isDone = false;
+    }
+
+    /**
+     * Checks if task description contains given keyword, case insensitive.
+     * 
+     * @param keyword Keyword to search for.
+     * @return True, if description contains given keyword. False otherwise.
+     */
+    public boolean hasKeyword(String keyword) {
```
ZhengLong3, line 52: I like that your boolean method sounds like a boolean.

src/main/java/cheese/storage/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Parses string from save file to task.
+     * 
+     * @param taskString String from save file to parse.
+     * @return Task parsed from string.
+     * @throws CheeseException If string from save file is corrupted.
+     */
+    private Task decodeStringToTask(String taskString) throws CheeseException {
```
ZhengLong3, line 105: I like how your methods are very descriptive in what they do. This can be seen in many other instances.

ZhengLong3 (commented):
The code looks very good to me.
All the methods are documented and the names are all very well chosen.
I also cannot find any coding standard violations.

src/main/java/cheese/task/Task.java
jasonqiu212: thank you ZhengLong for reviewing my code!

src/main/java/cheese/command/ByeCommand.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public class ByeCommand extends Command {
+    /**
+     * Executes operation to display goodbye message.
+     * 
+     * @param {@inheritDoc}
+     */
+    @Override
+    public void execute(TaskList taskList, Storage storage, Ui ui) {
+        ui.showGoodbye();
```
gunbux: Something that could be done, is that since it seems that you are only using `ui` to print certain strings, maybe you could use `ui` as a static class, that way you don't need to keep parsing it in as a variable, you could just do `Ui.showGoodbye()`

src/main/java/cheese/Cheese.java
```diff
@@ -0,0 +1,66 @@
...
+     * 
+     * @param filePath File path of save file.
+     */
+    public Cheese(String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            taskList = storage.load();
+        } catch (CheeseException e) {
+            ui.showError(e.getMessage());
```
gunbux, line 35: Although we're only supposed to comment on styling issues, maybe an alternative I'd like to suggest that is seems like you could probably use `Ui` as a static class, that way you don't need to keep parsing it into different parts of the code as an argument!

gunbux (commented):
Looks really great! Everything was very well named and organized, it was basically self documenting code!

src/main/java/cheese/data/TaskList.java
```diff
@@ -0,0 +1,151 @@
...
+    /**
+     * Validates if index given is within range of task list.
+     * 
+     * @param taskIndex Index of task to validate.
+     * @throws CheeseException If index is not in the range of task list.
+     */
+    private void validateTaskIndexInRange(int taskIndex) throws CheeseException {
+        if (taskIndex < 0 || taskIndex >= taskList.size()) {
+            throw new CheeseException("Item number is not in list range.");
+        }
```
gunbux: Really like how consistent your naming conventions are (i.e. `validate` for functions that throw errors), makes the code really easy to understand!

src/main/java/cheese/command/ByeCommand.java
jasonqiu212: that's a good idea! I'll refactor this for the Gui

src/main/java/cheese/data/TaskList.java
jasonqiu212: thanks for reviewing my code!

PR #456, opened 2022-08-25
[Wang Xinyi] iP

wxy1203:
### **DUKEPRO** :balloon:

> “Success is often achieved by those who don’t know that failure is inevitable.” – Coco Chanel 

Nikki is a little cute chatbot that can help you to remember the things in your daily life, it can,

- keep track of different event you have
- give ~~several~~ few responses to users

All you need to do is simply download this little chatbot from [here](https://github.com/wxy1203/ip) :cake:

1. it is **FREE**!!! :tada:
2. You will love it!!! :heart_eyes:

Its feature for now:

- [x] help you note down the tasks to do
- [x] manage deadline, event and todo list
- [x] find a certain task
- [x] auto save the list content
- [x] consolidate the total task

***
If you Java programmer, you can use it to practice Java too. Here's the `main` method: :battery:
```java
public class Main { 
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public void handleDelete(int index) throws DukeException {
+        Task taskToBeDeleted = tasks.deleteTask(index);
+        int size = tasks.getSize();
+        ui.printDelete(taskToBeDeleted, size);
+        storage.save(tasks);
+    }
+
+    /**
+     * Filters the list based on keyword
```
laxus2308: I think you are missing out on punctuation at the end of this line.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,98 @@
...
+                        String str = input.nextLine().trim();
+                        duke.find(str);
+                        break;
+
+                    //Default case: Not any of the tasks(aTodo, Deadline, Event) and hence, throws an Exception
+                    default:
+                        //To handle any extra words the user keyed in
+                        input.nextLine();
+                        throw new DukeException("OOPS! I'm sorry, but I don't know what that means :-(");
+
```
laxus2308, line 86: Perhaps there shouldn't be an indentation for case clauses according to the Java coding standard? I noticed the same issue at other places as well.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,53 @@
...
+ * A Task class that represents a task inputted by user.
+ */
+public abstract class Task {
+
+    /**
+     * Constructor for the Task Class.
+     *
+     * @param description Description of task.
+     */
+    protected String description;
```
laxus2308, line 13: Perhaps it would be better to put the javadocs for constructor just above the constructor method?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,69 @@
+package duke;
+
+import java.util.ArrayList;
+
+public class TaskList {
+
+    private ArrayList<Task> lst;
```
laxus2308, line 7: Perhaps a plural form such as tasks could be used to name the variable? Might it seem more intuitive that way?

laxus2308 (commented):
Overall, I like how clean your code is and the naming of method names. Just a few Java coding standards that have to be noted.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,175 @@
...
+        ui.bye();
+    }
+
+    /**
+     * Handles a Todo task inputted by user by calling on TaskList and Ui objects.
+     *
+     * @param tDescription A String of the description for the task.
+     * @throws DukeException
+     */
+    public void addTodo(String tDescription) throws DukeException {
```
ruiqi7: I like how you used verbs for the method name and the method name also explains clearly what the method does!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @param str A string representing the entire input user keyed in after "deadline".
+     */
+    public void addDeadline(String str) throws DukeException {
+        LocalDate date;
+        String dDescription;
+        try {
+            if (str.equals("")) {
+                throw new DukeException("OOPS! The description of a todo cannot be empty.");
+            }
```
ruiqi7: I like how you dealt with the unusual case as soon as it is detected and kept the main path un-indented as much as possible!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,175 @@
...
+     * @param str A string representing the entire input user keyed in after "event".
+     */
+    public void addEvent(String str) throws DukeException {
+        String eDescription;
+        String eAt;
+        if (str.equals("")) {
+            throw new DukeException("OOPS! The description of a todo cannot be empty.");
+        } else {
+            eDescription = str.substring(0, str.indexOf('/') - 1);
+            eAt = str.substring(str.indexOf('/') + 4);
```
ruiqi7: Perhaps use a named variable to represent the number 4 to make it clear what the number means or where it came from?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,98 @@
...
+                    case "find" :
+                        String str = input.nextLine().trim();
+                        duke.find(str);
+                        break;
+
+                    //Default case: Not any of the tasks(aTodo, Deadline, Event) and hence, throws an Exception
+                    default:
+                        //To handle any extra words the user keyed in
+                        input.nextLine();
+                        throw new DukeException("OOPS! I'm sorry, but I don't know what that means :-(");
```
ruiqi7: I like that you included a default branch to detect unusual cases!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,98 @@
...
+     * @param duke The duke object the user is operating on.
+     */
+    public Parser(Duke duke) {
+        this.duke = duke;
+    }
+
+    /**
+     * Parses the various inputs inputted by user.
+     */
+    public void start() {
```
ruiqi7: Perhaps it might be clearer if the method name was more descriptive? eg. `parseUserInputs()`

ruiqi7 (commented):
In general, I found your code easy to read and understand! There's just a few modifications you could consider to further increase its readability! 😃

src/main/java/duke/Duke.java
```diff
@@ -160,6 +160,7 @@ public String find(String string) {
         }
         return ui.printFind(filtered);
     }
+    // Test sample
 
     /**
      * Returns a goodbye message to the user.
```
Lan-Jingbo, line 166: I think you need to screenshot the whole chat box but now two pictures

src/main/java/duke/Duke.java
```diff
@@ -118,7 +119,7 @@ public String printList() {
      * @throws DukeException
      */
     public String handleMark(int index) throws DukeException {
-        
+        assert index > 0;
```
Lan-Jingbo, line 109:
```
assert(index > 0) will be clearer, easy to read
```

PR #459, opened 2022-08-25
[Nyamdavaa Amar] iP

nvmdava:
#Duke
- Bullet
- List
1. Numbered
2. List
```java
public class Main {
    }
}
```
- [x] Done
- [ ] Not Done
💯💯💯
`block quote`
[hyperlink ](https://github.com/nus-cs2103-AY2223S1/ip/pull/459)
`Inline code == true`
~~strikethrough~~
_italic_
**bold**


src/main/java/duke/Command.java
```diff
@@ -0,0 +1,8 @@
+package duke;
+
+/**
+ * Command interface to be passed onto Parser.
+ */
+interface Command {
+    void execute(String string) throws DukeException;
```
WR3nd3: Perhaps you can write a javadoc comment for this execute method and {@inherit} the overriding methods?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,130 @@
...
+        Storage storage = new Storage("data.txt");
+        storage.fillData(taskList);
+
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+
+        UserInterface.print("Hello from\n" + logo);
```
WR3nd3: Maybe it would be better abstraction if all the code/methods constructing messages to print is also encapsulated into the UserInterface class instead of only using it to hold the print method.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    }
+
+    /**
+     * Initialize Parser with built-in commands.
+     */
+    private void fillParser() {
+        parser.addCommand("todo", argument -> {
+            if (argument.equals("")) {
```
WR3nd3: The String methods isBlank() and isEmpty() could replace the equals("")

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,130 @@
...
+                throw DukeException.INVALIDARGUMENT;
+            }
+            Task item = Task.toDo(argument);
+            taskList.addTask(item);
+            UserInterface.print("added: " + item);
```
WR3nd3: Maybe it would be better abstraction if all the code/methods constructing messages to print is also encapsulated into the UserInterface class instead of only using it to hold the print method. This would also be useful to make the getResponse() method for the upcoming iP tasks simpler to implement.

src/main/java/duke/Duke.java
WR3nd3: This can apply to all parts of the code that use UserInterface.print(String) since it appears very frequently.

src/main/java/duke/Command.java
WR3nd3: Similarly, javadoc headers are missing for most public classes/methods. I believe those should be completed.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,130 @@
...
+            } catch (DukeException e) {
+                UserInterface.print(e.getMessage());
+            }
+        }
+    }
+
+    /**
+     * Initialize Parser with built-in commands.
+     */
+    private void fillParser() {
```
WR3nd3, line 47: Would it be better to define the behavior of the commands in the Parser class instead of defining them in the Duke class?

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,128 @@
+package duke;
+
+import java.time.LocalDate;
+
+abstract class Task {
+    private String description;
+    private boolean isDone = false;
+
+    private static class ToDo extends Task {
```
WR3nd3, line 9: Would it be better to define the subclasses of Task in their own class files?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,130 @@
...
+
+        parser.addCommand("event", argument -> {
+            if (argument.equals("")) {
+                throw DukeException.INVALIDARGUMENT;
+            }
+            Task item = Task.event(argument);
+            taskList.addTask(item);
+            UserInterface.print("added: " + item);
+            UserInterface.print("Now you have " + taskList.getSize() + " tasks in the list.");
+        });
```
gowribhat: Do you think code duplication can be avoided here? Perhaps a helper function can be used.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,128 @@
...
+        return new Deadline(arr[0], arr[1]);
+    }
+
+    public static Task event(String argument) throws DukeException {
+        String[] arr = argument.split("\\|", 2);
+        if (arr.length < 2) {
+            throw DukeException.INVALIDARGUMENT;
+        }
+        return new Event(arr[0], arr[1]);
+    }
```
gowribhat: Perhaps the method names can be made clearer to show what it does in verb form (eg 'createDeadline') 😄

src/test/java/duke/TaskTest.java
```diff
@@ -0,0 +1,32 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+public class TaskTest {
+    @Test
+    public void TaskTest() {
```
gowribhat, line 10: Maybe you could have separate tests instead of one test with multiple assertions?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Get the task at given position
+     *
+     * @param i position of the task.
+     * @return the task at position.
+     * @throws DukeException when the task position is non-existent.
+     */
+    public Task getTask(int i) throws DukeException {
```
gowribhat, line 52: Would it be possible to abstract this out? You could potentially have methods to mark/unmark in `TaskList` class so as to avoid having to pass `Task` objects around 😄

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,83 @@
+package duke;
+
+import java.util.ArrayList;
+
+/**
+ * List to keep tasks.
+ */
+class TaskList {
+
+    private ArrayList<Task> listOfTasks = new ArrayList<>();
```
gowribhat, line 10: Good naming through the use of plural!

gowribhat (commented):
Overall, code quality guidelines have been followed well. Only some minor changes are needed 😄

PR #460, opened 2022-08-25
[Debbie Wu] iP

debwy:
Your mind will be freed by Dave. Dave will...
- Communicate via text
- Be easy to pick up
- Be **mediocre** in terms of speed

If you feel like it, you can
1. _Download_ it from [here](https://se-education.org/guides/conventions/java/intermediate.html)
2. _That's all I think_

**FREE!**

Features:
- [x] Task management
- [ ] Deadline management
- [ ] Reminders

Here is a Java `main` method
`
public class Main {
        // Hello World
    }
}
`

```Java
public class Main {
        // Hello World
    }
}
```


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,177 @@
...
+                }
+            }
+        } catch (DukeException e) {
+            System.out.println("Error occurred during loading: " + e);
+        } catch (IOException e) {
+            System.out.print("Error occurred with saving/loading: " + e);
+        }
+    }
+
+    private void execute(Command c, String s) throws IOException, DukeTaskException {
```
therealdaofu: I really like this function, looks clean and easy to understand

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,177 @@
...
+                        temp.Mark();
+                        System.out.println("This task is now done: \n" + temp);
+                    }
+                    case "unmark":
+                        if (!isMultipleWords) {
+                            throw new DukeException("Index of task to unmark required");
+                        }
+                        int index = Integer.parseInt(restWord) - 1; //array starts from 0
+                        Task temp = log.get(index);
+                        temp.Unmark();
```
therealdaofu: It looks like you are not using this chunk of code anymore, maybe you could consider deleting it? or just copy it to a different file just in case

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public Pair<Duke.Command, String> parse(String s) throws DukeException, DukeTaskException{
+
+        boolean isMultipleWords = false;
+        String firstWord = "";
+        String restWord = "";
+
+        int indexOfSpace = s.indexOf(' ');
+        isMultipleWords = indexOfSpace > -1;
+        firstWord = s;
```
therealdaofu: Perhaps you can initialise both firstword and restWord at lines 9 & 10

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+        log = new ArrayList<>();
+        String s;
+        char first;
+        while ((s = reader.readLine()) != null) {
+            first = s.charAt(0);
+            parse(first, s);
+        }
+    }
+
+    private void parse(char c, String s) throws DukeException {
```
therealdaofu, line 98: Might be cleaner if you use switch case here

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public void setLog(ArrayList<Task> log) {
+        this.log = log;
+    }
+
+    public ArrayList<Task> getLog() {
+        return log;
+    }
+
+    public Task addTodo(String s) throws DukeTaskException {
```
therealdaofu, line 44: Is it possible to generalise addTodo etc. to maybe something like addTask(Task newTask)?

src/main/java/main.java
```diff
@@ -0,0 +1,7 @@
+import duke.Duke;
+
+public class main {
```
zlimez:
```suggestion
public class Main {
```
I really like that you extracted the entry point of the application into its own class. It is very clean. Hmm but perhaps the class name main could be capitalised to follow the coding standard?

zlimez (commented):
Small coding standard review otherwise fantastic

PR #465, opened 2022-08-26
[Hui Yi Lu] iP

huiyicodes:
# DUKE
Duke helps you
- manage your **todos**
- keep track of your **deadlines**
- remember your **events**

All you need to do is
1. [Download the jar file](https://github.com/pewggls/ip/releases/tag/A-Jar) to an empty folder
2. Open a command window in the same folder 
3. Run the command `java -jar duke.jar`

You can use it to practice Java too. Here's the main method:

```java
public class Main {
        Duke duke = new Duke("data", "duke.txt");
        duke.start();    
    }
}
```
⭐ Features: ⭐ 
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

> _"Duke is awesome"_ - John Doe


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+import duke.task.Todo;
+
+public class Parser {
+    private boolean bye;
+
+    public Parser() {
+        this.bye = false;
+    }
+
+    public boolean isBye() {
```
janelleljt: Could there have been a better name for this method? `isBye` is might not make sense to someone reading your code.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,79 @@
...
+        return this.bye;
+    }
+
+    public void parse(String input, TaskList tasks) throws DukeException {
+        if (input.equals("bye")) {
+            this.bye = true;
+        } else if (input.equals("list")) {
+            tasks.showList();
+        } else if (input.startsWith("mark")) {
+            if (input.length() < 6) {
```
janelleljt: We should try to minimize hardcoding numbers without meaning into our code. Check out the `replace` function, it might be beneficial to replace the string "mark" with an empty string so we can check for the task given after mark.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,79 @@
...
+public class Storage {
+    private File directory;
+    private File file;
+
+    public Storage(String dirName, String fileName) {
+        this.directory = new File(dirName);
+        this.file = new File(dirName + "/" + fileName);
+    }
+
+    public ArrayList<Task> getTasksFromDisk() throws FileNotFoundException, DukeException {
```
janelleljt, line 41: Could use a more concise function name like `loadTasks()`

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.ArrayList;
+public class TaskList {
+    private ArrayList<Task> list = new ArrayList<>();
+
+    public TaskList(ArrayList<Task> list) {
+        this.list = list;
+    }
+
+    public void addTask(Task task) {
+        if (task instanceof Todo) {
```
janelleljt, line 32: We might want to avoid using `instanceof`. Is there a better way to check the type of Task? (Applies to other places where `instanceof` is used)

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,32 @@
...
+public class Ui {
+    private final String GREETING = "Hello";
+    private final String BYE = "Goodbye";
+    private Parser parser;
+
+    public Ui() {
+        this.parser = new Parser();
+    }
+
+    public void doGreeting() {
```
janelleljt:
```suggestion
    public void printGreeting() {
```
Perhaps this function could have been named better? `doGreeting` does not really make grammatical sense. (Same for `doBye`)

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public void markDone() {
+        this.isDone = true;
+    }
+
+    public void markNotDone() {
+        this.isDone = false;
+    }
+
+    public String save() {
```
janelleljt, line 73: Could there have been better naming for this function? It does not technically save (kind of implies something related to storage).

janelleljt (commented):
Overall, did not note many glaring errors. Good job and jia you!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,79 @@
+package duke;
+
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+public class Parser {
+    private boolean bye;
```
jontmy: Hi! Also wanted to point out that by the Java standard coding standard, boolean variables/methods should be named to sound like booleans. Perhaps a name like `isBye` instead, or maybe `isTerminal` instead?

jontmy (commented):
Your code is very easy to read and follows the Java coding standard well!

Just wanted to add on to the existing review, some of your class methods have extra or no lines separating them, which you may wanna edit.

PR #467, opened 2022-08-26
[Eng Yu Siang] iP

yusiangeng:
## Hi

- one

1. two

```js
let a = 1;
```

- [ ] three
- [x] four

🙃 

> five

[link](https://github.com/yusiangeng)

six seven `eight(nine)` ten

*eleven* **twelve** ***thirteen*** ~~fourteen~~

yay


src/main/java/Duke.java
```diff
@@ -1,10 +1,126 @@
...
+                } else if (input.startsWith("unmark ")) {
+                    markTask(input, false);
+                } else if (input.startsWith("delete ")) {
+                    deleteTask(input);
+                } else if (input.startsWith("todo ")) {
+                    addTask(input, TaskType.TODO);
+                } else if (input.startsWith("deadline ")) {
+                    addTask(input, TaskType.DEADLINE);
+                } else if (input.startsWith("event ")) {
+                    addTask(input, TaskType.EVENT);
```
shenyih0ng: Since you have already initialize the values for each `TaskType` enum (seen below), perhaps it will be good to reuse that enum to do input checking instead.

This is so that you will not need to change the logic for input checking twice when a situation arises in the future

https://github.com/nus-cs2103-AY2223S1/ip/blob/2fd2ae5db207e3a413b396cadf268f0a5c8753ae/src/main/java/Duke.java#L56-L59

src/main/java/Duke.java
```diff
@@ -1,10 +1,126 @@
...
+                : "Please add a description and date for the %s";
+        if (splitInput.length < 2) {
+            throw new DukeException(String.format(errorMessage, type.string));
+        }
+        Task task;
+        if (type == TaskType.TODO) {
+            String desc = splitInput[1];
+            task = new Todo(desc);
+        } else {
+            splitInput = splitInput[1].split(type == TaskType.DEADLINE ? " /by " : " /at ");
```
shenyih0ng: I guess at this point there are only a few different type of `Tasks`, so there doesnt seem to be the need to componentize the parsing logic for each add task command (for example a dedicated parser for different type of different `addTasks`)

However, I would think it will be still be good practise to separate out the parsing logic so that it is also easier pin out the source of bugs if there are any. Additionally it also introduce better separation of concern, allowing us to better handling different command formats effectively.

src/main/java/Duke.java
```diff
@@ -1,10 +1,126 @@
...
+    private static void deleteTask(String input) throws DukeException {
+        String[] splitInput = input.split("delete ");
+        if (splitInput.length < 2) {
+            throw new DukeException("Please specify the number of the task to delete");
+        }
+        try {
+            int taskIdx = Integer.parseInt(splitInput[1]);
+            Task task = tasks.remove(taskIdx - 1);
+            System.out.printf("Noted. I've removed this task:\n  %s\n", task.toString());
+            System.out.printf("Now you have %d tasks in the list.\n", tasks.size());
```
shenyih0ng: Instead of handling the printing of UI responses from within each task action for example `deleteTask` in thie case, perhaps it would be better if there is a dedicated class/component that handles that. This is so that when more features are added or a GUI is incorporated,  it is easily to adapt and change to the new structure.

For example instead of having `void deleteTask`, it could be a `Response deletTask` where each action returns a structured `Response` output that is subsequently used by a `Ui` class/component to do the appropiate UI output.

text-ui-test/input.txt
```diff
@@ -0,0 +1,8 @@
+todo borrow book
+deadline return book /by Sunday
+event project meeting /at Mon 2-4pm
+list
+mark 1
+list
+unmark 1
+list
```
shenyih0ng, line 8: Some additional test cases you can consider:

1. Invalid comand formats for `deadline`, `event` and `mark`
    - e.g. `deadline /by Sunday`, `event project meeting /by Mon`, `mark a`
2. Tests for `delete` command 
3. What should be the expected output of `list` if there are no tasks yet in the `taskList`
4. Error messages when a invalid command is given
    - Test for the correct error message for commands like `mark a`
5. Tests for unsupported commands (e.g. `hello`)
6. For completeness sake, the expected output of `bye` command 😅

src/main/java/Deadline.java
```diff
@@ -0,0 +1,15 @@
+public class Deadline extends Task {
+
+    protected String by;
```
shenyih0ng: nit: This could be `final`

src/main/java/Duke.java
```diff
@@ -1,10 +1,126 @@
...
+
+        public final String string;
+
+        TaskType(String string) {
+            this.string = string;
+        }
+    }
+
+    private static void addTask(String input, TaskType type) throws DukeException {
+        String[] splitInput = input.split(type.string);
```
shenyih0ng: nit: Since it is a collection of strings, maybe a better variable name could be `inputTokens`?

shenyih0ng (commented):
Overall the coding standard is adhered to and the logic of the program is clear and easy to understand :) 

It would be great to include header comments and documentation later on. Additionally, you can also consider breaking it up to packages as well to better separate different components.

PR #470, opened 2022-08-26
[Ryan Tan] iP

Ryan-Tan00:
# Candice

- text-based
- easy to learn

All you need to do is,

1. download it from [here](../../).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

##

```java
public class Main {
    }
}
```


src/main/java/Action.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    public static Action actionParser(String input) throws UnknownActionException, EmptyTaskNameException,
+            InvalidFormattingException, EmptyTimingException, NumberFormatException {
+        String[] actionSplit = input.split(" ", 2);
+
+        String action = actionSplit[0];
+
+        switch (action) {
```
HowSuen: For `switch` statements, there should be no indentation for `case` and `default` clauses

src/main/java/Action.java
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                } else {
+                    throw new EmptyTaskNameException(action);
+                }
+            default:
+                throw new UnknownActionException();
+        }
+    }
+
+    public abstract void resolve(ArrayList<Task> taskList);
```
HowSuen: Since your sub classes are overriding this `resolve` method, perhaps it would be better to place this abstract method at the top so we can expect to see this method in the sub classes?

src/main/java/Action.java
```diff
@@ -0,0 +1,160 @@
...
+                newTask =  new TimedTask.Deadline(this.taskDescription, this.timing);
+            } else {
+                newTask = new TimedTask.Event(this.taskDescription, this.timing);
+            }
+
+            Task.addTask(newTask, taskList);
+        }
+    }
+
+    public static Action actionParser(String input) throws UnknownActionException, EmptyTaskNameException,
```
HowSuen: I like that you have this function to handle the commands given by the user. :) 
Maybe you can consider splitting the actions into separate functions and use abstraction to call them? Perhaps it would make the code for this function shorter and simpler to follow!

src/main/java/TimedTask.java
```diff
@@ -0,0 +1,30 @@
...
+        this.timing = timing;
+    }
+
+    public static class Deadline extends TimedTask {
+        public Deadline(String taskName, String timing) {
+            super(taskName, timing);
+        }
+
+        @Override
+        public String status() {
```
HowSuen: Would it be better if you used verbs to name methods like these, eg. `showStatus()` instead of `status()`?

HowSuen (commented):
Overall, I like the flow of logic you used and it was easy to read and understand.
You may consider creating separate class files for your subclasses in `Action` and `Task` and package them under `action` and `task` packages for instance, as there is an increment A-Packages where you organise your classes into packages. It also makes your code more compact and less verbose. I also like the personality of your Duke bot!

src/main/java/TimedTask.java
```diff
@@ -1,43 +1,93 @@
...
+                return "[D], " + finishedStatus + ", " + this.taskName + ", " + this.date
+                        + ", no time given\n";
+            } else {
+                return "[D], " + finishedStatus + ", " + this.taskName + ", " + this.date + ", "
+                        + this.eventStartTime + ", " + this.eventEndTime + "\n";
+            }
         }
-
     }
 }
```
Gavzzz: Overall code looks good and clean and should have no problem completing the IP

PR #474, opened 2022-08-26
[Benjamin Sim] iP

Benjamin-Sim:

- text-based
- easy to learn

All you need to do is,
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [x] Reminders (coming soon)

---

``` java
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+    
+    private static final String DEADLINE_LETTER = "D";
+
+
+    private static LocalDateTime parseDateAndTimeString(String dateAndTime) {
+
+        // dateAndTime is of the form "yyyy-mm-dd hh:mm"
+        // Therefore, the first token is the date and the second token is the time
+
+        // Split string around whitespaces
```
parth-io: I feel this comment is a bit redundant, as the code is self-explanatory

src/main/java/Deadline.java
```diff
@@ -0,0 +1,72 @@
...
+
+
+    public Deadline(String description, boolean isDone, LocalDateTime dateAndTimeObject) {
+        super(description, isDone);
+        this.dateAndTime = dateAndTimeObject;
+    }
+
+
+
+
```
parth-io: These spaces can be deleted

src/main/java/Duke.java
```diff
@@ -1,10 +1,340 @@
...
+
+    // TaskList object to store the user's tasks
+    private TaskList storedTasks;
+
+
+    // Constructor
+    public Duke() {
+        // TaskList is initialized in the initialize() method
+    }
+
```
parth-io: This can be safely deleted to make it less confusing. The default constructor is an empty constructor

src/main/java/Duke.java
```diff
@@ -1,10 +1,340 @@
...
+            return isDeadline ? new Deadline(description, dateAndTime) : new Event(description, dateAndTime);
+
+        
+        default:
+            return null;
+        }
+    }
+
+
+    public boolean isValidEventCommand(String[] commands) {
```
parth-io: The naming for this method can be `hasValidEventCommand`, as the current name reads more like one for a boolean variable

src/main/java/Event.java
```diff
@@ -0,0 +1,35 @@
...
+        this(description, false, dateAndTime);
+    }
+
+    public Event(String description, boolean isDone, String dateAndTime) {
+        super(description, isDone);
+        this.dateAndTime = dateAndTime;
+    }
+    
+
+    @Override
```
parth-io: I like the `@Override` annonations, makes it clear!

src/main/java/Duke.java
```diff
@@ -1,10 +1,340 @@
...
+        String result = String.format("Got it. I've added this task:\n%s\nNow you have %d tasks in the list.\n",
+                                        t, storedTasks.getSize());
+        System.out.println(result);
+    }
+
+
+    public Task createTask(String[] commands) throws Exception {
+        
+        String description = "";
+        boolean isDeadline = false;
```
parth-io: Might be good to make these private

parth-io (commented):
In general, I found the code very well structured and easy to follow. The use of fallthroughs in switch statements seems to go against recommended practices, but I don't think it is an issue as the codebase is still small.

Comments can also be deleted; there are many redundant comments. For example,
```java
  // Constructors
    public ToDo(String description) {
        super(description);
    }
```

I found the naming to be very good, and there were long, descriptive names where appropriate, so that was nice and easy to follow code logic.

src/main/java/Duke.java
Benjamin-Sim: I think the accessibility modifier doesn't matter for these variables since they are part of a method, not an instance or class attribute

src/main/java/Deadline.java
Benjamin-Sim: I write these comments to help myself easily remember the logic for my code. 😅
But good to know that it's self-explanatory for others

PR #476, opened 2022-08-26
[Bryan Yang] iP

TheSoggy:
# DukePro

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://nus-cs2103-ay2223s1.github.io/website/admin/ip-w4.html).
1. double-click it.
1. add your tasks.
1. let it manage your tasks for you 😉

And it is ****FREE****!

Features:

- [X] Managing tasks
- [ ] Reminders (coming soon)
-----------------------------------------

If you Java programmer, you can <mark>use</mark> it to practice Java too. Here's the `main` method:
```Java
public class Main {
        Application.launch(Main.class, args);
    }
}
```


src/main/java/IOParser.java
```diff
@@ -0,0 +1,22 @@
+import org.apache.commons.text.WordUtils;
+
+import java.util.Arrays;
+
+public class IOParser {
```
Devanshshah1309: Heyy, any reason the `IOParser` doesn't parse the user's input? I think parsing of user input is being done in `Duke` as of now, right?

src/main/java/Main.java
```diff
@@ -0,0 +1,10 @@
+public class Main {
```
Devanshshah1309: Any reason for not using the `main` method of `Duke` class?

src/main/java/Task.java
```diff
@@ -0,0 +1,100 @@
...
+     * Returns the string representation of a Task.
+     *
+     * @return The string representing the Task.
+     */
+    @Override
+    public String toString() {
+        return String.format("[%c][%s] %s", symbol, getStatusIcon(), description);
+    }
+
+    private static class TodoTask extends Task {
```
Devanshshah1309: Would it be better to have these as separate files rather than static nested classes?

src/main/java/Task.java
```diff
@@ -0,0 +1,100 @@
...
+        public TodoTask(Map<String, String> args) {
+            super('T', args.get("description"));
+        }
+    }
+
+    private static class DeadlineTask extends Task {
```
Devanshshah1309: Similarly for this too

Devanshshah1309 (commented):
Great work :) Clean and understandable code 👍

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+public class Storage {
+
+    private static final String taskListTxt = "duke.txt";
```
TYKCodes: Perhaps the name of your constant variable should be capitalised? Maybe something like:
```suggestion
    private static final String TASKLIST_TXT = "duke.txt";
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,207 @@
...
+import duke.command.Command;
+import duke.command.CommandType;
+import duke.command.DeleteTaskCommand;
+import duke.command.EmptyCommand;
+import duke.command.ListCommand;
+import duke.command.SetDoneCommand;
+import duke.task.DeadlineTask;
+import duke.task.EventTask;
+import duke.task.Task;
+import duke.task.TodoTask;
```
TYKCodes: It's great that you listed all of the classes that you are importing instead of using a wildcard

TYKCodes (commented):
Overall, your code looks clean and understandable to read. Some of your classes and its methods are missing some documentation, but otherwise the other descriptions are well done and descriptive enough to understand

PR #477, opened 2022-08-27
[Nathaniel Lukas] iP

dreammac3816547290:
# Duke
>"The shorter way to do many things is to only do one thing at a time." - Mozart

Duke helps you keep track of ToDo, Event, and Deadline tasks. It's
- **TEXT-BASED**
- _simple_
- **_powerful_**

All you need to do is
1. download the duke.jar file from [here](https://github.com/dreammac3816547290/ip/releases).
2. open the command prompt and navigate to the location of the downloaded file.
3. type java -jar duke.jar and enter.
4. use it however you want! :smiley:

Features:
- [x] Add ToDos, Deadlines, and Events.
- [ ] List all tasks.
- [ ] Delete tasks.
- [ ] Mark tasks as done or not done.

If you are a Java programmer, you can use it to practice Java too. Here's the `main` method
```java
    new Duke().start();
}
```


src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void markNotDone() {
+        isDone = false;
+    }
+
+    public String data() {
```
lpohsien: Perhaps the name for the method could be change to one that starts with a verb like `convertToData`?

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public class Task {
```
lpohsien: Perhaps this `Task` class and it's derived class could be put into a single package?

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+    protected boolean isDone;
+    protected char type;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+        type = 'T';
+    }
+
+    public String getStatusIcon() {
```
lpohsien: Maybe you can add some method headers for the public methods?

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+                    if (task.trim().isEmpty()) {
+                        throw new DukeException("☹ OOPS!!! The description of a todo cannot be empty.");
+                    }
+                    tasks.add(new Task(task));
+                    System.out.println("Got it. I've added this task:");
+                    System.out.println(tasks.get(tasks.size() - 1));
+                    System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+                } else if (s.equals("deadline")) {
+                    String rest = sc.nextLine();
+                    String[] split = rest.split("/");
```
lpohsien: Considering that the `split` is an array of strings, perhaps it could named using the plural convention? Another way to name it could be `tokens` which might be more informative?

lpohsien (commented):
Overall, I found the code easy to follow. The coding standard was consistent with only a few nits that might need fixing. Descriptive headers for public classes and methods could be included to allow for better documentation.

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+                    sc.nextLine();
+                } else {
+                    String rest = sc.nextLine();
+                    throw new DukeException("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+                }
+                modify(tasks);
+            } catch (DukeException e) {
+                System.out.println(e.getMessage());
+            }
+        }
```
Tan-Jin-Waye: Do consider the use of a switch statement as opposed to a large if-else block.

~~~java
switch (s) {
case "bye":
    // Code for exit
    break;
case "list":
    // Code for list print
   break;
...

~~~

src/main/java/Duke.java
```diff
@@ -1,10 +1,143 @@
...
+            } else if (split[0].equals("D")) {
+                tasks.add(new Deadline(split[2], split[3]));
+                if (split[1].equals("1")) {
+                    tasks.get(tasks.size() - 1).markAsDone();
+                }
+            } else if (split[0].equals("E")) {
+                tasks.add(new Event(split[2], split[3]));
+                if (split[1].equals("1")) {
+                    tasks.get(tasks.size() - 1).markAsDone();
+                }
```
Tan-Jin-Waye: Again, consider using a switch statement.
A switch fallthrough can be used to deal with the Deadline (line 115) and Event (line 120) cases elegantly.

~~~java
switch (split[0]) {
case "T":
    // Code for Todo
    break;
case "D":
   // Fallthrough
case "E":
   // Code for Deadline and Event
   break;
~~~

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    protected char type;
```
Tan-Jin-Waye: Nice use of protected, though realistically not necessary for description and isDone.

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+    public String toString() {
+        return "[" + type + "][" + getStatusIcon() + "]" + description;
+    }
+
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void markNotDone() {
+        isDone = false;
```
Tan-Jin-Waye: May want to include some code to check if the user is attempting to markAsDone an already done task, and same for markNotDone, and respond to user accordingly.

Tan-Jin-Waye (commented):
Code employed is sufficient to pass muster for Level 7/8, but better design choices need to be applied for later levels. I assume you're busy with other commitments at the moment, in which case all the best.

Jiayous.

PR #485, opened 2022-08-27
[Silas Yeo] iP

comicalromance:
# DukePro

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 🐱

And it is **FREE**!

## Features:
 
- [x] Managing tasks
- [ ] Reminders (coming soon)

If you Java programmer, you can use it to practice Java too. Here's the main method:

```java
public class Duke {
        new Duke(FILE_LOCATION, FOLDER_LOCATION).run();
    }
}
```


src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,15 @@
...
+
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+
+public abstract class Command {
+
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage);
+
+    public boolean isClose() {
```
tingkai-mai: I think the naming "isClose" might be a bit confusing, how about something like "isExit"?

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,20 @@
+package duke.command;
+
+import duke.Storage;
+import duke.TaskList;
+import duke.Ui;
+
+public class DeleteCommand extends Command {
+
+    int index;
```
tingkai-mai: Hmm, do you think this should be a private variable?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public void mark() {
+        isDone = true;
+    }
+
+    public void unmark() {
+            isDone = false;
+    }
+
+    public String getStatusIcon() {
```
tingkai-mai, line 42: Great job separating this into its own function! This makes checking for "X" very clear.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,39 @@
+package duke.task;
+
+public class Task {
+    enum Tag {
```
tingkai-mai, line 7: The enum values for the Tags aren't immediately clear, especially when they're used within the code. (e.g. I wasn't too sure what Tag.T meant at first.) Might you want to consider renaming this?

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Parses raw input and returns corresponding command.
+     *
+     * @param input Raw input string.
+     * @return Command to be executed.
+     * @throws Exception  If there is an error in parsing.
+     */
+    public static Command parse(String input) throws Exception {
```
tingkai-mai, line 62: Is it better to throw a DukeException here?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    /**
+     * Saves text containing task data into save file.
+     *
+     * @param tasks A list of strings containing lines to save.
+     * @throws Exception if save file cannot be written.
+     */
+    public void save(TaskList tasks) throws Exception {
```
tingkai-mai: Since you threw a DukeException for `load`, you might want to consider throwing that for `save` as well!

tingkai-mai (commented):
LGTM! I think I would suggest renaming some of the enum values to make it clearer, and perhaps throw/catch exceptions using DukeException rather than Exception.

src/main/java/duke/command/Command.java
comicalromance: I agree that exit might cause less ambiguity for users!

src/main/java/duke/task/Task.java
comicalromance: Thanks for the compliment!

src/main/java/duke/task/Task.java
comicalromance: I understand that the tags may be a bit confusing, but this method is used to reduce overhead when calling the toString method which simply prints tag. However, this definitely can be resolved by modifying the tag names and adding a new method to return a one char tag for printing.

src/main/java/duke/Parser.java
comicalromance: While it would be great to throw a DukeException for consistency, some parts of the code can throw other exceptions that I have not yet handled as an DukeException, so I can only throw Exception as a quick fix for now.

src/main/java/duke/Storage.java
comicalromance: Good spot! Will handle the save function with DukeException in mind.

src/main/java/duke/command/DeleteCommand.java
comicalromance: Good spot! I should specify the type of access modifier explicitly instead of using the default and making people wonder if it is deliberately or accidentally default.

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,58 @@
...
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+public class DialogBox extends HBox {
+
+    @FXML
+    private Label dialog;
+    @FXML
+    private ImageView displayPicture;
```
rexong, line 23: Perhap it is better to name is as a noun? example `pictureDiplayer`.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Creates a new Duke Object with the default folder and file location.
+     */
+    public Duke() {
+        this(FILE_LOCATION, FOLDER_LOCATION);
+    }
+
+    public boolean getLoaded() {
```
rexong, line 41: Maybe the method name can be changed to `isLoaded` instead?

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.junit.jupiter.api.Test;
+
+import duke.command.EventCommand;
+import duke.command.ExitCommand;
+import duke.command.MarkCommand;
+import duke.command.UnmarkCommand;
```
rexong, line 11: I like how your import statements are in order. 👍

rexong (commented):
Your code looks good. Most of it is following the coding standard! Just some of the minor suggestions regarding your method and attributes naming.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,85 @@
...
+            Scene scene = new Scene(ap);
+            stage.setScene(scene);
+            fxmlLoader.<MainWindow>getController().setDuke(duke);
+            mainWindow = fxmlLoader.<MainWindow>getController();
+            mainWindow.setDuke(duke);
+            printGreetings();
+            if (duke.getLoaded()) {
+                print("File successfully loaded!");
+            } else {
+                print("Error parsing load file");
```
tanhl2000, line 40: This can be put under Exception handling instead as it is an error message that should show with an exception. DukeException could also be expanded upon to have multiple more specific exceptions.

tanhl2000 (commented):
Overall great job with the code! Coding standards mostly followed as well.

PR #502, opened 2022-08-29
[Bag Devesh Kumar] iP

tensaida:
# Drake
> She said, "Do you love me?"
> I tell her, "Only partly"
> I only love my bed and
> managing my user's tasks, I'm sorry
 – Drake, _God's Plan_ 

Drake the type of guy to free your mind of having to remember things you need to do. It's,

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://github.com/tensaida/ip/releases/download/v0.1/duke.jar).
2. double-click him.
3. add your tasks.
4. let him manage your tasks for you 😉

And it is **FREE**!

## Features

- [x] Managing tasks
- [x]  Managing deadlines 
- [x]  Reminders 

If you're a Java programmer, you can use it to practice Java too. Here's the main method:

```java
public class Main {
        new Drake().run();
    }
}
```


src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Objects;
+import java.util.Scanner;
+
+
+public class Drake {
+
+    public static void main(String[] args) throws DrakeException, IOException {
+
```
wweqg: Hi I notice the exceptions here are being thrown instead of being handled. Would this terminate the program under some circumstances ? 🤔

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+        Scanner sc = new Scanner(System.in);
+        String command = sc.nextLine();
+
+        ArrayList<Task> list = new ArrayList<>();
+
+        File taskFile = new File("data/tasks.txt");
+        if (taskFile.createNewFile()) {
+            System.out.println("Task file created: " + taskFile.getName());
+        } else {
+            System.out.println("Task file already exists.");
```
wweqg: I think it is fine to not prompt this reminder. `System.out.println("Task file already exists.")`
Assuming the file is there, and every time after the user entered command, he/she will see 
```
......
.....
(User enetered command)
Task file already exists.
........
```
I guess this will left user to wonder what does this reminder even mean 😕

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+                        }
+                        list.add(task);
+                        break;
+                    case "E":
+                        Event event = new Event(taskParts[2], taskParts[3]);
+                        if (taskParts[1].equals("X")) {
+                            event.markAsDone();
+                        }
+                        list.add(event);
+                        break;
```
wweqg: Would it be better if you named these numbers into constants ? 🤔  For example:
```
final int TASK_TYPE = 0;
....
switch (taskParts[TASK_TYPE])
....
```

Also, there should not be any indentation for case. For example switch method would look like:
```
switch(number) {
case 1:
    ...
    break;
case2:
    ..
    break;
.....
```

wweqg (commented):
Overall, I feel that your program is pretty good, the structure is there and especially I like the way how you customize the response from Drake. 😄  

However just some minor issues here. I notice that you write everything in your main method, which I guess, is not a good practice if you would like to build extension (include extra features). I guess you actually had a sense of it already as `A-MoreOOP` is one of this IP requirements to met.

Yeah so I think once you abstracted your code in main method into methods from multiple classes, your program would be kind of flawless (based on my understanding). 👀 

I understand that this might take some time (personally I took one whole day to settle this OOP requirement), yup so if you need any help can always drop me a message. Also you could check out IP code from others on our IP dashboard to get a sense on how to organize different classes. 👍

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+            String firstWord;
+            if (command.equals("list")) {
+                firstWord = "list";
+            } else if (firstSpace == -1) {
+                throw new UnknownCommandException();
+            } else {
+                firstWord = command.substring(0, firstSpace);
+            }
+            System.out.println("------------------------------------------------------");
+            switch (firstWord) {
```
SilasTSL: `switch` and `case` should not be indented according to Java coding standard

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+                    System.out.println("I've added this task:");
+                    list.add(new Event(commands[0], commands[1].trim()));
+                    System.out.println(list.get(list.size() - 1));
+                    System.out.println("You now have " + list.size() + " tasks in the list");
+                    break;
+                }
+                case "delete": {
+                    String[] commands = command.split(" ");
+                    try {
+                        int taskNo = Integer.parseInt(commands[1]);
```
SilasTSL: Maybe it's clearer to name this variable as taskNumber instead of taskNo? I think this variable appears in other places too!

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+                    if (commands.length == 1)
+                        throw new IncompatibleCommandException("A deadline task without a deadline?");
+                    System.out.println("I've added this task:");
+                    list.add(new Deadline(commands[0], commands[1].trim()));
+                    System.out.println(list.get(list.size() - 1));
+                    System.out.println("You now have " + list.size() + " tasks in the list");
+                    break;
+                }
+                case "event": {
+                    String filtered = command.substring("event ".length());
```
SilasTSL: I think `filtered` could maybe be more specific?

SilasTSL (commented):
I think more OOP could be done. I love your funny console outputs!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public void writeToFile(FileWriter writer) throws IOException {
+        writer.write(String.format("D;%s;%s;%s\n", getStatusIcon(), description, by));
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by.format(DateTimeFormatter.ofPattern("dd MMM yyyy")) + ")";
+    }
+}
```
kavantan: Small issue, but I believe your file is missing a newline at the bottom. I noticed the same issue in several other files too.

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Objects;
+import java.util.Scanner;
+
+
+public class Drake {
+
+    public static void main(String[] args) throws DrakeException, IOException {
```
kavantan: I feel that your main method is slightly lengthy, perhaps you could consider refactoring some of the functionalities into other methods/classes?

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+                    break;
+                }
+                default:
+                    throw new UnknownCommandException();
+            }
+            System.out.println("------------------------------------------------------");
+            command = sc.nextLine();
+        }
+        taskFile.delete();
+        File updatedTaskFile = new File("data/tasks.txt");
```
kavantan: I noticed there are quite a few instances of repeated variables being used such as 'data/tasks.txt' and the long chain of hyphens '-----'. Have you considered declaring them as global constants instead?

src/main/java/Task.java
```diff
@@ -0,0 +1,33 @@
+import java.io.FileWriter;
+import java.io.IOException;
```
kavantan: I like that you followed the coding standards for your import statements 👍

src/main/java/Task.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void unmarkAsDone() {
```
kavantan: Minor issue, but have you considered renaming this method to markAsUndone() instead? I feel that unmarkAsDone might potentially leave some room for ambiguity.

kavantan (commented):
Overall, I find the code relatively easy to read and understand. Although as mentioned in the specific comments, perhaps you might want to add JavaDocs whenever possible and possibly refactor your code into subclasses/methods, to make your code even more readable. Apart from that, there are minor violations of the code quality guidelines and coding standard conventions, but I'm sure those can be rectified very easily. All the best!

src/main/java/Drake.java
linuschancs: Indeed! This may also cause file errors when your program terminates before your file is closed.

src/main/java/Drake.java
linuschancs: Agreed! Perhaps considering changing these numbers into constants could improve readability and also reduce the possibility of an invalid entry into the variable!

src/main/java/Event.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Event extends Task {
+    protected final LocalDate at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.at = LocalDate.parse(at);
```
linuschancs: A possible consideration would be to pass the entire String into the Event class and throw the errors from this class to be handled in the main class. This applies for your others tasks as well.

Also a possible scenario that occurs with the current implementation is what if users input an invalid formatted date/ random input as the date? What will happen when the input is parsed?

src/main/java/Drake.java
linuschancs: Indeed! Perhaps you could consider thinking of the program in terms of multiple tools working together to make it work for example the interpreter that makes sense of the user's commands, the UI that involves what the user interacts with, the task list to store tasks etc.

src/main/java/Drake.java
```diff
@@ -0,0 +1,183 @@
...
+                        if (taskParts[1].equals("X")) {
+                            deadline.markAsDone();
+                        }
+                        list.add(deadline);
+                        break;
+                    case "T":
+                        Task task = new Task(taskParts[2]);
+                        if (taskParts[1].equals("X")) {
+                            task.markAsDone();
+                        }
```
linuschancs: You could perhaps consider abstracting this into your Task class!

linuschancs (commented):
Overall, your code for the most part was easy to understand and followed coding convention, perhaps just for the main method a consideration would be to abstract out the logic into more classes and methods rather than handling all the logic there! There were a few minor coding standard violations such as indentations however nothing too serious!

PR #507, opened 2022-08-29
[Tin Jingyao] iP

tin-jy:

text-based
easy to learn
~~FAST~~ *SUPER FAST* to use
All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
+import java.util.ArrayList;
+import java.util.Locale;
+import java.util.Scanner;
+
 public class Duke {
```
edricyeo: Perhaps apply OOP to split the Duke class into different smaller classes?

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
+
+    private static final ArrayList<Task> LIST_OF_THINGS = new ArrayList<>();
+
+    private abstract static class Task {
```
edricyeo: Maybe you could create Task as a separate class, instead of nesting it in the Duke class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
+            public String toString() {
+                if (isCompleted()) {
+                    return String.format("[D][X] %s (By: %s)", getTaskName(), deadline);
+                } else {
+                    return String.format("[D][ ] %s (By: %s)", getTaskName(), deadline);
+                }
+            }
+        }
+
+        private static class EventTask extends Task {
```
edricyeo: I like how you named the different Tasks clearly! 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
+
+    public static boolean isUnmarkCommand(String str) {
+        if (str.length() < 8) {
+            return false;
+        }
+        if (!str.startsWith("unmark")) {
+            return false;
+        }
+        try {
+            Integer.parseInt(str.substring(7));
```
edricyeo: Perhaps you could consider using str.split(" ") rather than str.substring(7). May improve reusability of code instead of having to change the number according to the length of the command?

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
+import java.util.ArrayList;
+import java.util.Locale;
+import java.util.Scanner;
+
 public class Duke {
+
+    private static final ArrayList<Task> LIST_OF_THINGS = new ArrayList<>();
```
edricyeo: Maybe you could consider extracting this out as a separate TaskList class, which would handle adding tasks, etc.

edricyeo (commented):
Generally good effort!
I like the clear naming of classes, variables and methods.
I think generally you could split your code into different classes, and include JavaDocs for your classes and methods!

src/main/java/Duke.java
domoberzin: Could split the class into various classes, i.e. TaskList, Event, Deadline, instead of putting them into one class

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
+
+    private abstract static class Task {
+        private final String task_name;
```
domoberzin: Should be camelcase formatting for variables

src/main/java/Duke.java
domoberzin: Could also add JavaDocs for documentation purposes

domoberzin (commented):
Coding standard is generally adhered to w.r.t naming, but could extract out the different classes, as well as add JavaDocs. :)

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,5 @@
+package duke;
+
+public class Command {
+
+}
```
DarrenCsAcc: I am curious why is this empty?

PR #530, opened 2022-09-03
[Gavin Cho] iP

Gavzzz:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
+
+    protected String by;
```
rycs2812: I think it might be better if the String had a more descriptive name, perhaps something like "date" or "deadline"?

src/main/java/Duke.java
```diff
@@ -1,10 +1,130 @@
...
+
+        Task[] tasks = new Task[100];
+        int counter = 0;
+        ArrayList<Task> collection = new ArrayList<>();
+
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+
+
+
```
rycs2812: I think it might be better if there was only a single blank line here

src/main/java/Duke.java
```diff
@@ -1,10 +1,130 @@
...
+        while (!input.equals("bye")) {
+            if (input.equals("list")) {
+                System.out.println("___________________________________");
+                for (int i = 0; i < counter; i++) {
+                    System.out.println( (i+1) + "."
```
rycs2812: Seems like there's an extra whitespace on this line, might need to remove it to follow the coding standards

src/main/java/Duke.java
```diff
@@ -1,10 +1,130 @@
...
+                            + collection.get(i).toString());
+                }
+                System.out.println("___________________________________");
+                input = sc.next();
+
+            }
+
+            else if (input.equals("mark")) {
+                int number = sc.nextInt();
+                collection.get(number-1).mark();
```
rycs2812: Similar issue on missing/extra whitespaces, noticed this issue in several other places too, might have to address these to follow the coding standards

src/main/java/Duke.java
```diff
@@ -1,10 +1,130 @@
...
+
+
+    public static void main(String[] args) throws DukeException {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("___________________________________");
```
rycs2812: Perhaps the line break could be stored in a constant since it is used multiple times in the code

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
rycs2812: Good job on following the coding standards for naming boolean variables👍

rycs2812 (commented):
Apart from the styling issues, I think the code is generally pretty okay. As long as all the increments are implemented eventually, I think the project will turn out fine!

src/main/java/Duke.java
```diff
@@ -1,10 +1,201 @@
...
+            System.out.println("File has been created: " + file.getName());
+        } else {
+            System.out.println("File already exists.");
+            Scanner reader = new Scanner(file);
+            while (reader.hasNext()) {
+                reader.useDelimiter(" \\| ");
+                Constants constant = Constants.valueOf(reader.next().toUpperCase());
+                Task task;
+                switch (constant) {
+                    case TODO:
```
ssyap98: There should not be indentation for the case clauses. It should look something like this
`switch (constant) {`
 `case TODO:`

src/main/java/Event.java
```diff
@@ -0,0 +1,22 @@
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class Event extends Task{
+
+    protected String at;
```
ssyap98: Maybe you can change the variable name from "at" to "eventTime" so that it sounds more meaningful?

ssyap98 (commented):
Overall code looks good! Just have to make some minor changes to follow the java coding standard!

